PR #27, opened 2015-03-08
WIP: charwidth function

### LINKED ISSUE #2: add charwidth property
As discussed in JuliaLang/julia#6939, the `wcwidth` function is broken on many operating systems.   When we import the Unicode data, it would be good to add another field to our database in order to store the character width, so that we can provide an up-to-date character-width function.

stevengj:
This PR fixes #2 by following [the algorithm suggested](http://nbviewer.ipython.org/gist/jiahao/07e8b08bf6d8671e9734) by @jiahao: it uses character widths from GNU unifont, plus a few exceptions from UAX 11 and some control characters.

One difference from @jiahao's script is that I divide unifont advance widths by 512 *before* incorporating the UAX 11 data, whereas @jiahao's notebook did the division afterwards, which seemed like a bug(??).

Also, I'm not quite sure what we should do about non-printable characters.  Should we return -1 for non-printable characters like wcwidth?  If so, what algorithm for "printable" should we use?  @jiahao's notebook had one algorithm, but Julia's isprint function subsequently adopted a different algorithm.  Right now, I am returning 0; my feeling is that if the caller wants to know some more specific notion of "printable," they can always inspect the unicode category code provided by utf8proc.


data/charwidths.jl
```diff
@@ -0,0 +1,151 @@
...
+        if state==:seekchar         #StartChar: nonmarkingreturn
+            if contains(line, "StartChar: ")
+                codepoint = nothing
+                width = nothing
+                state = :readdata
+            end
+        elseif state==:readdata #Encoding: 65538 -1 2, Width: 1024
+            contains(line, "Encoding:") && (codepoint = int(split(line)[3]))
+            contains(line, "Width:") && (width = int(split(line)[2]))
+            if codepoint!=nothing && width!=nothing && codepoint >= 0
```
stevengj: @jiahao, I added the codepoint>=0 check here since the sfd file seems to have some codepoint -1 entries in the beginning.

data/charwidths.jl
```diff
@@ -0,0 +1,151 @@
...
+#
+# Requires Julia (obviously) and FontForge.
+
+#############################################################################
+# Widths from GNU Unifont
+
+universion="7.0.06"
+for fontfile in ["unifont-$universion", "unifont_upper-$universion"]
+    isfile("$fontfile.ttf") || download("http://unifoundry.com/pub/unifont-$universion/font-builds/$fontfile.ttf", "$fontfile.ttf")
+    isfile("$fontfile.sfd") || run(`fontforge -lang=ff -c "Open(\"$fontfile.ttf\");Save(\"$fontfile.sfd\");Quit(0);"`)
```
stevengj: Note that -lang=ff seems to be needed for recent versions of fontforge, since they changed the -c option to accept Python scripts by default rather than FontForge's own scripting language.

nalimilan:
Maybe return `0` rather than `-1` for non-printable characters? That way, people won't subtract 1 to the total width of a string without noticing if it accidentally contains such a character. Ideally they should check and remove them, but it is generally easier to compute the width without modifying the string, and skip non-printable chars only in the code that actually does the printing.

stevengj:
@nalimilan, currently I'm returning `0` for non-printables.  The main argument for returning `-1` would be to make it a drop-in replacement for `wcwidth`.   Of course, we could just add a `utf8proc_wcwidth` that did this, but we still have to decide what "printable" means.   (It would also be good to understand the original rationale for this behavior of `wcwidth`.)

stevengj:
I just noticed the [Python-based wcwidth repo](https://github.com/jquast/wcwidth) on github; it would be good to compare to their behavior.

stevengj:
@jiahao, I noticed that your algorithm seems to assign a nonzero width to [category Mn (non-spacing combining characters)](http://www.fileformat.info/info/unicode/category/Mn/list.htm).   e.g. [U+0300 (combining grave accent)](http://www.fileformat.info/info/unicode/char/0300/index.htm) is assigned a width of 1.   That seems weird to me...shouldn't we assign these a width of zero, since in practice their width is subsumed by the width of the character they are combined with?

stevengj:
@staticfloat, Travis seems to be failing occasionally because the Unifont downloads are stalling.  Is there some problem with large (few MB) downloads in Travis?  If so, how do the julia Travis builds work?

staticfloat:
We've got some [caching infrastructure](https://github.com/staticfloat/cache.julialang.org) that we use to make sure we get consistent results with as high uptime as possible.  The basic formula is to change your request URLs to something like `https://cache.e.ip.saba.us/http://www.unicode.org/Public/UNIDATA/UnicodeData.txt`.  The caching server is fairly picky about what it will cache, so I could add this to [the whitelist](https://github.com/staticfloat/cache.julialang.org/blob/a468d7c6a556a9449272e50f81a3a75e23795ff2/cache.py#L18-L49) and allow you to query the caching server for those files.

stevengj:
@staticfloat, it would be great if you added the unifont `.ttf` files we are using to the filter (with wildcards so that we can change versions without problems), sincd those seem to be the problem cases.

staticfloat:
Alright, it should be live now.  Try changing the URLs and let me know.

stevengj:
Thanks @staticfloat, it seems to work on my machine; hopefully Travis will succeed now.

stevengj:
(Another weird thing: why does the Travis badge for the `master` branch say "build:failing" when it is only an unmerged PR that has failed?)

stevengj:
Ah, it looks like I need to pass `--location` to `curl` so that it follows redirects.

tkelman:
> (Another weird thing: why does the Travis badge for the  master  branch say "build:failing" when it is only an unmerged PR that has failed?)

I think the markdown for the badge link is currently set to "most recent build for project," it needs some extra arguments to be specifically for master

stevengj:
@jiahao, it looks like the problem is the UAX11 code.   That assigns a width "1" to category "A" (ambiguous), which includes the Mn non-spacing combining marks.   Like I mentioned above, I'm not sure how your UAX11 code worked at all in your notebook, because you divided the widths by 512 *after* the UAX11 code, which would have set all of those widths to zero.  Maybe I just shouldn't check UAX11 at all, or should check that before Unifont (so that Unifont can override UAX11)?

stevengj:
Okay, just pushed another update.  Now Unifont takes precedence over UAX11, and I added a check.  Unifont includes PUA glyphs from the ConScript registry (e.g. Klingon, Elvish, etcetera), but I'm skeptical that we should return nonzero widths for those as they aren't part of the Unicode standard, so I changed the algorithm to return zero for these (category Co).

It now agrees with the MacOS 10.10.2 `wcwidth` function in all cases where `wcwidth` returns a positive result.  (`wcwidth` returns `-1` for many codepoints that we recognize as printable, however.)

stevengj:
@staticfloat, I'm still getting Travis timeouts on the downloads.  Am I using the caching server incorrectly somehow?

data/charwidths.jl
```diff
@@ -0,0 +1,158 @@
...
+#By definition, should be narrow = width of 1 en space
+#0x00202f ' ' category: Zs name: NARROW NO-BREAK SPACE/
+CharWidths[0x202f]=1
+
+#By definition, should be wide = width of 1 em space
+#0x002001 ' ' category: Zs name: EM QUAD/
+#0x002003 ' ' category: Zs name: EM SPACE/
+CharWidths[0x2001]=2
+CharWidths[0x2003]=2
+
```
stevengj: @jiahao, in a monospaced font, like in a terminal, aren't em and en the same?

jiahao:
@stevengj thanks for taking the lead on actually making this happen.

Dividing by 512: probably a silly bug. I may have tinkered with the notebook since the version I posted.

Category Mn having width 1: I think you're right, they should have width 0. The oversight did occur to me at some point but I don't think I updated the gist since I realized the mistake.

The 'ambiguous' category was difficult for me to figure out. Some characters like ❶ (U+02776 DINGBAT NEGATIVE CIRCLED DIGIT ONE) I have only ever seen in printed Chinese literature, where it is typeset as fullwidth. Others like ¸ (U+00B8 CEDILLA, a spacing character not to be confused with ̧ U+0327 COMBINING CEDILLA) seem silly to treat as ambiguous. I guess if ❶ were to ever show up in non-East Asian text the rule is to assign it as narrow? [UAX 11 §4.2 Ambiguous Characters](http://www.unicode.org/reports/tr11/tr11-28.html#Ambiguous) is really, well, ambiguous. Or one can simply adopt the recommendation of §5 which suggests that always pretending that they are narrow is a good default:

> When processing or displaying data... Ambiguous characters behave like wide or narrow characters depending on the context (language tag, script identification, associated font, source of data, or explicit markup; all can provide the context). If the context cannot be established reliably, they should be treated as narrow characters by default.

em vs en: The 'em' is defined to be equal to the current point size and the 'en' is defined to be half an 'em'. Given that the bounding box for Unifont was 1024 x 512 for most of the ordinary narrow characters in Fontforge, I assigned anything defined to be 'en'-sized as width 1 and anything 'em'-sized as width 2.

staticfloat:
@stevengj I think I forgot to enable a certain fix on the caching server, it's been updated now, and I've re-run the jobs which seemed to work better.

stevengj:
Thanks!   @jiahao, can you take a look at the patch?  There are some changes from your algorithm.

stevengj:
@jiahao, Unifont gives width=1 for four "W" codepoints U+2329, U+232a, U+fe33, U+fe34 and one "F" codepoint U+ff3d according to `EastAsianWidth.txt`.  I guess we want UAX11 to "win" in these cases, right?

jiahao:
Yes I recall seeing some discrepancies in the Unifont W characters having width 1. I don't remember the details but I think the Unifont bounding boxes were quite clearly erroneous. (Some of them are clearly fullwidth Chinese characters.)

jiahao:
Letting Unifont decide neutral and ambiguous seems better to me too, especially since anyone who actually wants monospaced Unicode will have little choice other than Unifont in practice.

PR #28, opened 2015-03-08
WIP: Minimal cmake build script

### LINKED ISSUE #23: cmake script
It would be nice to include a `cmake` script (a `CMakeLists.txt` file) to make life easier for Windows users.   Since utf8proc is so simple, we can easily maintain both this and a `Makefile`.

tkelman:
The `disallow_intree_builds()` snippet is thanks to Tracy Wadleigh. Tested on Linux, will check with MSVC/MinGW next.

will close #23

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,8 +8,8 @@
 *.dll
 *.dylib
 *.dSYM
-*.txt
 *.out
+data/*.txt
 bench/bench
 bench/icu
 bench/unistring
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,3 +8,11 @@ script:
     - make prefix=`pwd`/local install
     - make check
     - make utf8proc_data.c.new && (diff utf8proc_data.c.new utf8proc_data.c > /dev/null)
+    - mkdir build_static
+    - cd build_static
+    - cmake .. -DCMAKE_VERBOSE_MAKEFILE=ON
+    - make
+    - mkdir ../build_shared
+    - cd ../build_shared
+    - cmake .. -DCMAKE_VERBOSE_MAKEFILE=ON -DBUILD_SHARED_LIBS=ON
+    - make
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -6,6 +6,11 @@ disallow_intree_builds()
 
 project (utf8proc C)
 
+# Be sure to also update these in Makefile!
+set(SO_MAJOR 1)
+set(SO_MINOR 2)
+set(SO_PATCH 0)
+
 add_definitions (
   -DUTF8PROC_EXPORTS
 )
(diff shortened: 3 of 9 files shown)


tkelman:
Tested on MSVC and MinGW (`-G"MSYS Makefiles"`) too. Builds either static or shared library, defaults to static unless you ask for `-DBUILD_SHARED_LIBS=ON`. Can still be spruced up for building the tests, adding an install target, etc.

.gitignore
```diff
@@ -8,7 +8,6 @@
 *.dll
 *.dylib
 *.dSYM
-*.txt
```
stevengj: Add `data/*.txt` (or similar) back in, to exclude the Unicode data files that are downloaded there.

stevengj:
Don't move `lump.txt` into `data/`, since it is documentation and not data.  Maybe rename it to `lump.md` and make a Markdown-formatted version.  (Also need to change mentions of `lump.txt` in the README and `utf8proc.[ch]` accordingly.)

stevengj:
Also, can you add a test to the `.travis.yml` file?  (And we should test the Windows version with AppVeyor.)

stevengj:
Also should have a `VERSION` and `SOVERSION`, right?

tkelman:
> Also should have a `VERSION` and `SOVERSION`, right?

Yeah. Will be a little annoying to have to remember to keep the Makefile and CmakeLists synchronized here, won't it?

kmsquire:
Why not just have CMakeLists.txt?

tkelman:
If we make CMakeLists.txt do everything that the current Makefile does, sure. Should we just duplicate the info until that point?

kmsquire:
I would think so, assuming @stevengj is okay with that plan.

stevengj:
In the long term, we can think about moving over to `cmake` wholesale.  In the short term (utf8proc 1.2), I'd prefer to keep both and just duplicate the info.

tkelman:
Will https://github.com/tkelman/libutf8proc/commit/ce053beeb63a805fb22e4b825259a9acc7de00f4 suffice for now? See the travis/appveyor logs for what that does to the output file names.

stevengj:
@tkelman, it looks like this is calling `-Wl,-soname,libutf8proc.so.1.2.0`, which is wrong; the soname should be `-Wl,-soname,libutf8proc.so.1`.   `libutf8proc.so.1.2.0` is the library's "real name" — see [this HOWTO](http://tldp.org/HOWTO/Program-Library-HOWTO/shared-libraries.html).

tkelman:
Only using `${SO_MAJOR}` seems to work. The "real name" is getting set to `libutf8proc.so.1.2-dev` but I guess that's alright for now?

nalimilan:
The `-dev` suffix looks weird for a .so file name. Can't you get rid of it?

stevengj:
@tkelman, it sounds like `cmake` wants both `SOVERSION` and `VERSION` to refer to the ABI version, not the API version.  hence `SOVERSION` should be `1` and `VERSION` should be `1.2`.

tkelman:
Okay, I did that. Now if only Travis would run.

Issue #26, opened 2015-03-08
Documentation 


stevengj:
Currently, the API is documented in the header file. Maybe we should use Doxygen or similar to generate nicely formatted docs from the header?

What tool has the best GitHub integration?

PR #29, opened 2015-03-18
RFC: Use doxygen for generating API docs

jonas:
Hi,

To address issue #26, I've started working on converting the header file so doxygen can be used to generate API docs. It's not complete yet, hence the "RFC" prefix. I would however like to know if the general direction is good. A couple of things to consider:

* The comments are quite invasive, especially for the `#define` statements. If it is preferable, the comments could go before and simply reference the symbols they document.
* I'm using markdown as much as possible to keep the comments readable.
* Comments follow the initial formatting of utf8proc (recent comments didn't follow this):
```
/** Some title.
 *  Note the two spaces before the main text in the comment.
 */
```
* `@return` and `@param` are used, although it requires that the comments are restructured.

You can see the docs here: http://jonas.nitro.dk/utf8proc/html/index.html
With the important part being: http://jonas.nitro.dk/utf8proc/html/utf8proc_8h.html

I've also briefly looked at whether to use Sphinx for generating docs through the [breathe](https://github.com/michaeljones/breathe) plugin. This might allow us to create better looking docs. I don't know if that would be of interest, else I also found a way to use bootstrap with doxygen to make it look more modern.

[ I'm considering using utf8proc in [tig](https://github.com/jonas/tig) so I am very excited that you have taken over maintaining this great library. ]


jonas:
Forgot to mention that I didn't integrate anything into the build system for now.

tkelman:
Cool! This is probably a case of "he who implements it gets to decide what tool to use," but this looks not bad at all. Would breathe allow us to host everything on readthedocs? Judging by http://cppformat.readthedocs.org/en/stable/ as an example, that does look a lot nicer than the doxygen default.

jiahao:
I'm quite used to doxygen, so I wouldn't mind.

I would make a few small changes:

```patch
@@ -242,7 +242,7 @@ TCL_SUBST              =
 # members will be omitted, etc.
 # The default value is: NO.
 
-OPTIMIZE_OUTPUT_FOR_C  = NO
+OPTIMIZE_OUTPUT_FOR_C  = YES
 
 # Set the OPTIMIZE_OUTPUT_JAVA tag to YES if your project consists of Java or
 # Python sources only. Doxygen will then generate output that is more tailored
@@ -793,7 +793,7 @@ RECURSIVE              = NO
 # Note that relative paths are relative to the directory from which doxygen is
 # run.
 
-EXCLUDE                =
+EXCLUDE                = NEWS.md lump.md
 
 # The EXCLUDE_SYMLINKS tag can be used to select whether or not files or
 # directories that are symbolic links (a Unix file system feature) are excluded
```

jonas:
@tkelman Yes, as I understand it, breathe allows doc hosting on readthedocs. I didn't actually try to set it up yet. In any case, Doxygen is required to "lift" the docs from source.

@jiahao Thanks. Was not sure whether the external files made sense.

stevengj:
Maybe the `#define` statements should be replaced by `enum`s?

utf8proc.h
```diff
@@ -268,166 +308,214 @@ typedef struct utf8proc_property_struct {
...
- *  NLF2LS:  converts LF, CRLF, CR and NEL into LS
- *  NLF2PS:  converts LF, CRLF, CR and NEL into PS
- *  NLF2LF:  converts LF, CRLF, CR and NEL into LF
- *  STRIPCC: strips or converts all non-affected control characters
- *  COMPOSE: tries to combine decomposed characters into composite
- *           characters
- *  STABLE:  prohibits combining characters which would violate
- *           the unicode versioning stability
+ *
+ *  @param buffer the sequence of unicode characters to reencode.
```
stevengj: the UTF-32 unicode codepoints to re-encode

utf8proc.h
```diff
@@ -268,166 +308,214 @@ typedef struct utf8proc_property_struct {
...
+ *  @param length the length of the sequence.
```
stevengj: the length (in codepoints) of the buffer

utf8proc.h
```diff
@@ -268,166 +308,214 @@ typedef struct utf8proc_property_struct {
...
+ *  @par
  *  If the character is not existent a pointer to a special struct is
- *  returned, where 'category' is 0 (UTF8PROC_CATEGORY_CN).
+ *  returned, where `category` is 0 (@ref UTF8PROC_CATEGORY_CN).
  */
+DLLEXPORT const utf8proc_property_t *utf8proc_get_property(int32_t uc);
 
+/** Decompose a unicode char into an array of unicode chars.
+ *
+ *  @param uc the unicode char.
```
stevengj: Let's use "codepoint" consistently (instead of "code point" or "char" or "character")

utf8proc.h
```diff
@@ -45,8 +43,11 @@
  *    (i.e. UTF-8 encoded UTF-16 surrogates)
  *  - support for korean hangul characters
  *  Unicode Version 7.0.0 is supported.
```
tkelman: we should add either a newline or another `-` here, otherwise this is showing up on the same line as "support for korean hangul characters"

utf8proc.h
```diff
@@ -268,166 +308,214 @@ typedef struct utf8proc_property_struct {
...
- *  returned, where 'category' is 0 (UTF8PROC_CATEGORY_CN).
+ *  returned, where `category` is 0 (@ref UTF8PROC_CATEGORY_CN).
  */
+DLLEXPORT const utf8proc_property_t *utf8proc_get_property(int32_t uc);
 
+/** Decompose a unicode char into an array of unicode chars.
+ *
+ *  @param uc the unicode char.
+ *  @param dst the destination buffer.
+ *  @param bufsize the size of the destination buffer.
```
stevengj: Maybe use "length" for length of an array `T*` (in units of `T`), rather than "size" which could mean size in bytes.

jonas:
Thanks for the comments. I've tried to address the suggestions in the additional commits. I kept the changes separate so that they can easily be included or dropped in a final squashed commit.

Regarding the enum's it really cleans up the docs a lot. I had to use a typedef for the categories to not conflict with the utf8proc_category method and decided to also use a typedef for the options to keep things consistent.

utf8proc.h
```diff
@@ -178,65 +201,82 @@ typedef struct utf8proc_property_struct {
...
+  UTF8PROC_CATEGORY_ZL = 24, /**< Separator, line */
+  UTF8PROC_CATEGORY_ZP = 25, /**< Separator, paragraph */
+  UTF8PROC_CATEGORY_CC = 26, /**< Other, control */
+  UTF8PROC_CATEGORY_CF = 27, /**< Other, format */
+  UTF8PROC_CATEGORY_CS = 28, /**< Other, surrogate */
+  UTF8PROC_CATEGORY_CO = 29, /**< Other, private use */
+} utf8proc_category_t;
+
+/** Bidirectional character classes. */
+enum utf8proc_bidi_class {
```
stevengj: `typedef enum { .... } utf8proc_bidi_class_t` for consistency?

utf8proc.h
```diff
@@ -268,166 +312,213 @@ typedef struct utf8proc_property_struct {
...
- *  The maximum number of bytes read is 'strlen', unless 'strlen' is
- *  negative.
- *  If a valid unicode char could be read, it is stored in the variable
- *  being pointed to by 'dst', otherwise that variable will be set to -1.
- *  In case of success the number of bytes read is returned, otherwise a
- *  negative error code is returned.
+/**
+ * Reads a single codepoint from the UTF-8 sequence being pointed to by 'str'.
+ * The maximum number of bytes read is 'strlen', unless 'strlen' is
+ * negative.
```
stevengj: negative (in which case up to 4 bytes are read).

stevengj:
Great, can you update your doc page so that we can see what it looks like?

jonas:
@stevengj It's updated.

.gitignore
```diff
@@ -12,6 +12,7 @@
 data/*.txt
 data/*.ttf
 data/*.sfd
+/docs/
```
stevengj: What is the `docs` directory for?  The generated docs?  I'm not sure we should distribute those in the tarball as opposed to just posting them.

nalimilan: Distributions prefer shipping docs with packages rather than only online. That's required some work in Julia, so if it works here please keep them.

jonas: It goes with this line in `Doxyfile`:
```
+OUTPUT_DIRECTORY       = docs
```
It contains `docs/html`, `docs/man` etc.

stevengj:
Looks good, can you squash the commits for merging?

jonas:
@stevengj Squashed. I also added 'Fix #26' to the commit title.

stevengj:
Now, the next steps:

* How do we autogenerate the docs in for the github-created tarballs?

* How do we autogenerate the docs for readthedocs or similar, so that the online docs are auto-updated in sync with github?

nalimilan:
No idea about the second point, but as regards GitHub-generated tarballs, I think they're best avoided. Usually, projects have a `make dist` or `make distcheck` target which builds everything that is needed an creates the tarball. A good `make distcheck` tries to build the resulting tarball and run the tests from it, so that you can be sure you didn't forget a file in the release, or some silly thing like this.

I'm not saying this is very high priority, but that's what I see as a good practice.

jonas:
Once we've added a `make docs` and `make dist`, we can look into
configuring Travis to build release assets and upload them to GitHub:
http://docs.travis-ci.com/user/deployment/releases/

As for posting docs to readthedocs I can look into it. It looks we need to
install a webhook to notify ReadTheDocs and then add some configuration
files to run Doxygen and Sphinx:
https://docs.readthedocs.org/en/latest/webhooks.html#github
https://breathe.readthedocs.org/en/latest/readthedocs.html

PR #35, opened 2015-05-09
Fix #34 handle 66 Unicode non-characters and surrogates correctly

### LINKED ISSUE #34: utf8proc does not correctly handle the 66 Unicode "noncharacters"
`utf8proc` considers the 66 Unicode "noncharacters" to not be valid, however the Unicode standard specifically says that they are valid code points, and need to be handled correctly in conforming software.

ScottPJones:
This change also improves the performance of the functions utf8proc_iterate and utf8proc_codepoint_valid, and fixes surrogate handling in utf8proc_encode_char.

### DIFF
--- a/test/graphemetest.c
+++ b/test/graphemetest.c
@@ -6,6 +6,7 @@ int main(int argc, char **argv)
     size_t bufsize = 0;
     FILE *f = argc > 1 ? fopen(argv[1], "r") : NULL;
     utf8proc_uint8_t src[1024];
+    int len;
     
     check(f != NULL, "error opening GraphemeBreakTest.txt");
     while (getline(&buf, &bufsize, f) > 0) {
@@ -29,9 +30,10 @@ int main(int argc, char **argv)
             else if (buf[bi] == '#') { /* start of comments */
                 break;
             }
-            else { /* hex-encoded codepoint */
-                bi += encode((char*) (src + si), buf + bi) - 1;
+	    else { /* hex-encoded codepoint */
+                len = encode((char*) (src + si), buf + bi) - 1;
                 while (src[si]) ++si; /* advance to NUL termination */
+                bi += len;
             }
         }
         if (si && src[si-1] == '/')
@@ -59,7 +61,7 @@ int main(int argc, char **argv)
                        utf8proc_errmsg(glen));
                  for (i = 0; i <= glen; ++i)
                       if (g[i] == 0xff)
-                           g[i] = '/'; /* easier-to-read output (/ is not in test strings) */
+                          g[i] = '/'; /* easier-to-read output (/ is not in test strings) */
                  check(!strcmp((char*)g, (char*)src),
                        "grapheme mismatch: \"%s\" instead of \"%s\"", (char*)g, (char*)src);
             }
--- a/test/tests.h
+++ b/test/tests.h
@@ -33,10 +33,11 @@ size_t skipspaces(const char *buf, size_t i)
    separated by whitespace, and terminated by any character not in
    [0-9a-fA-F] or whitespace, then stores the corresponding utf8 string
    in dest, returning the number of bytes read from buf */
+utf8proc_ssize_t unsafe_encode_char(utf8proc_int32_t uc, utf8proc_uint8_t *dst);
 size_t encode(char *dest, const char *buf)
 {
      size_t i = 0, j, d = 0;
-     do {
+     for (;;) {
           int c;
           i = skipspaces(buf, i);
           for (j=i; buf[j] && strchr("0123456789abcdef", tolower(buf[j])); ++j)
@@ -45,9 +46,9 @@ size_t encode(char *dest, const char *buf)
                dest[d] = 0; /* NUL-terminate destination string */
                return i + 1;
           }
-          check(sscanf(buf + i, "%x", &c) == 1, "invalid hex input %s", buf+i);
+          check(sscanf(buf + i, "%x", (unsigned int *)&c) == 1, "invalid hex input %s", buf+i);
           i = j; /* skip to char after hex input */
-          d += utf8proc_encode_char(c, (utf8proc_uint8_t *) (dest + d));
-     } while (1);
+          d += unsafe_encode_char(c, (utf8proc_uint8_t *) (dest + d));
+     }
 }
 
--- a/utf8proc.c
+++ b/utf8proc.c
@@ -108,51 +108,57 @@ UTF8PROC_DLLEXPORT const char *utf8proc_errmsg(utf8proc_ssize_t errcode) {
   }
 }
 
+#define utf_cont(ch)  (((ch) & 0xc0) == 0x80)
 UTF8PROC_DLLEXPORT utf8proc_ssize_t utf8proc_iterate(
   const utf8proc_uint8_t *str, utf8proc_ssize_t strlen, utf8proc_int32_t *dst
 ) {
-  int length;
-  int i;
-  utf8proc_int32_t uc = -1;
+  utf8proc_uint32_t uc;
+  const utf8proc_uint8_t *end;
+
   *dst = -1;
   if (!strlen) return 0;
-  length = utf8proc_utf8class[str[0]];
-  if (!length) return UTF8PROC_ERROR_INVALIDUTF8;
-  if (strlen >= 0 && length > strlen) return UTF8PROC_ERROR_INVALIDUTF8;
-  for (i=1; i<length; i++) {
-    if ((str[i] & 0xC0) != 0x80) return UTF8PROC_ERROR_INVALIDUTF8;
+  end = str + ((strlen < 0) ? 4 : strlen);
+  uc = *str++;
+  if (uc < 0x80) {
+    *dst = uc;
+    return 1;
   }
-  switch (length) {
-    case 1:
-    uc = str[0];
-    break;
-    case 2:
-    uc = ((str[0] & 0x1F) <<  6) + (str[1] & 0x3F);
-    if (uc < 0x80) uc = -1;
-    break;
-    case 3:
-    uc = ((str[0] & 0x0F) << 12) + ((str[1] & 0x3F) <<  6)
-      + (str[2] & 0x3F);
-    if (uc < 0x800 || (uc >= 0xD800 && uc < 0xE000) ||
-      (uc >= 0xFDD0 && uc < 0xFDF0)) uc = -1;
-    break;
-    case 4:
-    uc = ((str[0] & 0x07) << 18) + ((str[1] & 0x3F) << 12)
-      + ((str[2] & 0x3F) <<  6) + (str[3] & 0x3F);
-    if (uc < 0x10000 || uc >= 0x110000) uc = -1;
-    break;
+  // Must be between 0xc2 and 0xf4 inclusive to be valid
+  if ((uc - 0xc2) > (0xf4-0xc2)) return UTF8PROC_ERROR_INVALIDUTF8;
+  if (uc < 0xe0) {         // 2-byte sequence
+     // Must have valid continuation character
+     if (!utf_cont(*str)) return UTF8PROC_ERROR_INVALIDUTF8;
+     *dst = ((uc & 0x1f)<<6) | (*str & 0x3f);
+     return 2;
   }
-  if (uc < 0 || ((uc & 0xFFFF) >= 0xFFFE))
-    return UTF8PROC_ERROR_INVALIDUTF8;
-  *dst = uc;
-  return length;
+  if (uc < 0xf0) {        // 3-byte sequence
+     if ((str + 1 >= end) || !utf_cont(*str) || !utf_cont(str[1]))
+        return UTF8PROC_ERROR_INVALIDUTF8;
+     // Check for surrogate chars
+     if (uc == 0xed && *str > 0x9f)
+         return UTF8PROC_ERROR_INVALIDUTF8;
+     uc = ((uc & 0xf)<<12) | ((*str & 0x3f)<<6) | (str[1] & 0x3f);
+     if (uc < 0x800)
+         return UTF8PROC_ERROR_INVALIDUTF8;
+     *dst = uc;
+     return 3;
+  }
+  // 4-byte sequence
+  // Must have 3 valid continuation characters
+  if ((str + 2 >= end) || !utf_cont(*str) || !utf_cont(str[1]) || !utf_cont(str[2]))
+     return UTF8PROC_ERROR_INVALIDUTF8;
+  // Make sure in correct range (0x10000 - 0x10ffff)
+  if (uc == 0xf0) {
+    if (*str < 0x90) return UTF8PROC_ERROR_INVALIDUTF8;
+  } else if (uc == 0xf4) {
+    if (*str > 0x8f) return UTF8PROC_ERROR_INVALIDUTF8;
+  }
+  *dst = ((uc & 7)<<18) | ((*str & 0x3f)<<12) | ((str[1] & 0x3f)<<6) | (str[2] & 0x3f);
+  return 4;
 }
 
 UTF8PROC_DLLEXPORT utf8proc_bool utf8proc_codepoint_valid(utf8proc_int32_t uc) {
-  if (uc < 0 || uc >= 0x110000 ||
-    ((uc & 0xFFFF) >= 0xFFFE) || (uc >= 0xD800 && uc < 0xE000) ||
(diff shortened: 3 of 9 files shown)


pao:
`make check` failed on Travis, possible encoding problem? (Fancy that in a Unicode library.)

pao:
Or possibly it's now testing non-characters because the definition changed.

ScottPJones:
The errors are something to do with graphemes, which I didn't think my changes would even affect.
There are also a number of warnings about a wcwidth function, on both gcc and clang, that I didn't touch :sad:

nalimilan:
Given that both Travis builds failed (AppVeyor does not run the tests), it doesn't look like a random failure. It doesn't look unlikely to me that modifying an essential function like `utf8proc_encode_char` could change the behaviour of tests. Even if you didn't change these functions, they probably call the modified code indirectly.

ScottPJones:
@nalimilan What makes me wonder a bit, is all the warning messages in the travis compilation...
I'll dig into it though

stevengj:
Would also be good to add a regression test to the `tests/` directory.

ScottPJones:
Note: I figured out why the grapheme test failed... but it doesn't make me very happy about utf8proc.
I managed to work around it, but basically, the tests *depended* on the utf8 encoding being broken!
(it encoded some values as 0xFF and 0xFE, which it then used as markers, but left that bad encoding
for *all* callers to the utf8proc_encode_char function).
@stevengj This already has fairly complete tests..., are you talking about additional tests in the julia tests unit tests?

stevengj:
Yes, the `CHARBOUND` option was documented as inserting `0xFF` bytes as delimiters.    I'd be happy to deprecate this functionality, since I find `utf8proc_grapheme_break` to be much more useful (because it doesn't require making a copy of the string etcetera just to detect grapheme breaks).   

Alternatively, we could change it to insert NUL bytes as delimiters, but it would be good to change the API (e.g. to `UTF8PROC_GRAPHEME`) in that case to force a breaking change for code relying on this feature.

By test case, I mean creating an additional test program that checks for regressions on handling of these non-characters and surrogates.

test/graphemetest.c
```diff
@@ -49,7 +51,7 @@ int main(int argc, char **argv)
                 else
                     i++;
             }
-            glen = utf8proc_map(utf8, j, &g, UTF8PROC_CHARBOUND);
+            glen = utf8proc_map(utf8, j, &g, UTF8PROC_CHARBOUND | UTF8PROC_OLDENCODE);
```
stevengj: Shouldn't `UTF8PROC_CHARBOUND` imply `UTF8PROC_OLDENCODE`?  Why have a separate flag?

ScottPJones: I'd missed that (about CHARBOUND...).
OK, I'll try to get some testing in... (in my copious spare time! :grinning:)
Would it be possible to have extra testing as a follow-up PR?

stevengj: In general, it's good practice to merge patches only given regression tests and (where needed) documentation.

ScottPJones: Yes, true... I'll add some further testing, and make sure it gets all of the edge cases that were broken before... there are no API changes though.

utf8proc.c
```diff
@@ -186,6 +186,37 @@ UTF8PROC_DLLEXPORT utf8proc_ssize_t utf8proc_encode_char(utf8proc_int32_t uc, ut
 }
 
 /* internal "unsafe" version that does not check whether uc is in range */
```
stevengj: Any non-`static` functions should be prefixed with `utf8proc_` in order to prevent namespace pollution.

utf8proc.c
```diff
@@ -165,12 +168,9 @@ UTF8PROC_DLLEXPORT utf8proc_ssize_t utf8proc_encode_char(utf8proc_int32_t uc, ut
...
     return 2;
-  } else if (uc == 0xFFFF) {
-    dst[0] = 0xFF;
-    return 1;
-  } else if (uc == 0xFFFE) {
-    dst[0] = 0xFE;
-    return 1;
+  } else if (((utf8proc_uint32_t)uc)-0xd800 < 0x800) {
+    // Surrogates not valid code point, can't be encoded!
+    return 0;
```
stevengj: I wonder if it makes more sense to be permissive about this...

ScottPJones: I don't think so, those are totally bad conversions...

ScottPJones: Or, did you mean for surrogates, not the 0xffff ones I removed?

stevengj: I meant for surrogates, which can appear from incorrect conversions of UTF-16 to UTF-8 (or Java-like "modified UTF-8"); I'm wondering what the tradeoffs here are for being permissive in the encoding step, so that we least provide the option of allowing such nonsense.

(We definitely want to detect them and throw an error *somewhere*, e.g. if `UTF8PROC_REJECTNA` is requested, but in that case it is optional.)

ScottPJones: OK, my take on this, is that for *validity* checking (unless we have some optional flags to make it relaxed... default strict), it should always be strict.  I think only conversion functions should potentially be relaxed, and then only for the two semi-standard variants, Modified UTF-8 (i.e. like Java), and CESU-8, like
Oracle, MySQL, and others (and that the conversion functions in Base only *produce* 100% correct Unicode... if somebody wants to produce something like Modified UTF-8 or CESU-8, they can easily write that in a package...)

utf8proc.c
ScottPJones: OK, that came about because I first did the changes based on what I found in my julia directory... only later did I see that the code really lived in JuliaLang/utf8proc instead of JuliaLang/julia, and missed fixing those... Sorry!

utf8proc.c
stevengj: Except this isn't a validity checking step per se (I agree that the utf8proc_map function should check for surrogates), this is an encoding function.  The question is whether to barf if the user specifically requests to encode a surrogate char.

I just want to be careful about changing the behavior here.  This is a behavior change in an existing API, and whether it is a bug fix is a matter of perspective.

ScottPJones: OK, I see your point... (it doesn't matter to me as much anymore, because I now would like to just cut Julia's dependence on utf8proc.c anyway... make it pure Julia code if possible, with maybe just a C data file... [but not the one from utf8proc_data.c]).   I'll change the surrogate part back.

utf8proc.c
ScottPJones: Not sure I understand your question about test functions (may be total lack of sleep and jet lag...).  Could you explain further?  Thanks!

test/tests.h
```diff
@@ -45,9 +46,9 @@ size_t encode(char *dest, const char *buf)
                dest[d] = 0; /* NUL-terminate destination string */
                return i + 1;
           }
-          check(sscanf(buf + i, "%x", &c) == 1, "invalid hex input %s", buf+i);
+          check(sscanf(buf + i, "%x", (unsigned int *)&c) == 1, "invalid hex input %s", buf+i);
           i = j; /* skip to char after hex input */
-          d += utf8proc_encode_char(c, (utf8proc_uint8_t *) (dest + d));
-     } while (1);
+          d += encode_char(c, (utf8proc_uint8_t *) (dest + d));
```
tkelman: need to put utf8proc_ prefix back

ScottPJones: It's actually, not putting `utf8proc_` prefix back, but rather adding `utf8proc_unsafe_`.
The version that starts with just `utf8proc_` would not work correctly...  My jetlagged brain missed that test case :sad:

utf8proc.c
stevengj: `unsafe_encode_char` is used in `tests.h`

ScottPJones: Ah, that was because of that graphemetest, that used the broken version to map 0xffff -> 0xff, and 0xfffe -> 0xfe.

stevengj: The `tests.h` file is only used for parsing the input test files from Unicode consortium into UTF-8 strings.  The `GraphemeBreakTest.txt` file from the Unicode consortium contains (unpaired) surrogate "code points", presumably for testing UTF-16 code.   The test program encodes these into (invalid) UTF-8, and then `utf8proc_map` (correctly) signals a `UTF8PROC_ERROR_INVALIDUTF8` error.

The graheme test doesn't use 0xFFFF or 0xFFFE, as far as I can tell.  I'm not sure why these cases were included in the `encode` function.  (Maybe some kind of BOM thing?  Not that BOM is needed or recommended for UTF-8 data.)

ScottPJones:
@stevengj Could you also please take a look at my new PR #39 here?
I added some reasonably extensive testing of both the `codepoint_valid` and `iterate` functions...
I did this as a separate PR, so you could easily see the issues I found with the current utf8proc.c code...
Thanks greatly!

utf8proc.c
```diff
@@ -186,7 +188,38 @@ UTF8PROC_DLLEXPORT utf8proc_ssize_t utf8proc_encode_char(utf8proc_int32_t uc, ut
...
+      dst[0] = 0xF0 + (uc >> 18);
+      dst[1] = 0x80 + ((uc >> 12) & 0x3F);
+      dst[2] = 0x80 + ((uc >> 6) & 0x3F);
+      dst[3] = 0x80 + (uc & 0x3F);
+      return 4;
+   } else return 0;
+}
+
+/* internal "unsafe" version that does not check whether uc is in range */
+static const utf8proc_property_t *utf8proc_unsafe_get_property(utf8proc_int32_t uc) {
```
stevengj: Since this is a `static` function, there is no need to put the `utf8proc_` prefix.   It's good to preserve the convention that exported (non-`static`) functions get the prefix, whereas non-exported functions have no prefix.

stevengj:
Once 39 is merged, can you update the tests to also test for noncharacter validity and include the updated tests in this PR?

ScottPJones:
#39 already does test for noncharacter validity...  If you run it now, it will show you all the problems you have if you *don't* merge this one :grinning:

utf8proc.c
ScottPJones: I added those because I mistakenly thought that's what you were asking me to do... I'll go back and  put those back... (this is what you get for programming on 3 days of only 2 hours of sleep!)

stevengj:
@ScottPJones, then it should be merged into this PR (and squashed/rebased).

stevengj:
Oh, you should also update the `.gitignore` file with the names of the new test programs (so that `git status` will ignore them).

ScottPJones:
But then, isn't it harder to show issues that haven't been fixed yet?
Actually, that's something that I think is a problem with the way the current unit testing works...
I think that you should be able to independently of fixes, be able to add unit tests for any problems that have been found... and the unit test framework should check for regressions on any particular test, while saving and displaying information about those that fail, instead of terminating the testing process...
(I worked this way for many years, it was *very* effective in not losing track of issues in the product, while showing where progress was being made in cleaning up old issues)

ScottPJones:
Oh, didn't know about .gitignore... Thanks!

stevengj:
Committing tests that fail is not really compatible with the Travis/Github workflow.  If you don't merge the PRs, then we are forced to either (a) merge 39 first, in which case you break the build (tests passing/failing is binary on Github) or (b) merge 35 first, in which case we are merging before testing.

timholy:
> I think that you should be able to independently of fixes, be able to add unit tests for any problems that have been found.

I agree this would be a major improvement. There have been quite a few times where I've had to omit a test that should have been there simply because some other bug made it fail. It would be really nice to have a "todo list" of failing tests.

That said, certainly our current framework doesn't support this.

ScottPJones:
@stevengj The way the unit tests would work within the Travis/GitHub framework is fairly simple... A "failure" for Travis would only be when there is a *regression* on a test that passed previously...
New tests show what the behavior is supposed to be, and the ones that don't match, get counted up,
and so you'd see something like:
```
Unicode Tests:
     Line 30, 'isvalid(UTF8String, b"\xff") -> false' failed, returned true
     .... 
  1000 passed, 50 failed, 3.53 seconds
```

What do you think?

stevengj:
@ScottPJones, that's reasonable, but since you have to build this facility "manually" I wouldn't bother on a project as small as utf8proc.  For Julia, it might be worthwhile to extend `Base.Test` with a `@todo` test (for something that is expected to fail) or similar.

ScottPJones:
@stevengj Thanks for all the review!   Hope I got the move of #39 here done correctly...

test/tests.h
```diff
@@ -33,10 +33,11 @@ size_t skipspaces(const char *buf, size_t i)
    separated by whitespace, and terminated by any character not in
    [0-9a-fA-F] or whitespace, then stores the corresponding utf8 string
    in dest, returning the number of bytes read from buf */
+utf8proc_ssize_t unsafe_encode_char(utf8proc_int32_t uc, utf8proc_uint8_t *dst);
 size_t encode(char *dest, const char *buf)
 {
      size_t i = 0, j, d = 0;
-     do {
+     for (;;) {
```
stevengj: BTW, is there some substantive reason you changed this?  Don't most compilers generate exactly the same code for `for (;;) {...}` and `do {...} while(1)`?

Usually, it's best to avoid purely stylistic changes in patches (unless you have a patch that is purely for a global style change); on the contrary, one strives for a patch that matches the surrounding code style.

ScottPJones:
@stevengj  You're quite right, but maybe this should be raised as an issue?  To improve unit testing for Julia?
We actually had a system that kept track of the results of previous unit tests, in CSV files... so it wasn't really "manual", but 1) I didn't build the unit testing framework, 2) it's not the sort of programming I really want to do :grinning:, but I know it can be done and is very very useful, and would be good if *somebody* smart could implement it... (sort of intermediate beginner project? or Summer of Code?)

tkelman:
There are lots of nice small usable test frameworks out there that are better than what we have now - we could just pick one and start using it. I like the one that libgit2 uses for example: https://github.com/vmg/clar

test/tests.h
ScottPJones: I don't even remember doing that... that sort of thing happens on autopilot... I date myself :grinning:, but when I started, not all C compilers optimized `while(1)` away!

ScottPJones: Anyway, do I need to change that back and resquash?

ScottPJones:
@tkelman  I'm pretty ignorant of a lot of what's out there in OSS land... (too many years with "not invented here" mentality where I worked... unfortunately I wasn't able to change that...)
Does vmg/clar have the capabilities that I described, of checking for regressions on each individual test?

utf8proc.c
```diff
@@ -165,12 +171,8 @@ UTF8PROC_DLLEXPORT utf8proc_ssize_t utf8proc_encode_char(utf8proc_int32_t uc, ut
...
     dst[1] = 0x80 + (uc & 0x3F);
     return 2;
-  } else if (uc == 0xFFFF) {
-    dst[0] = 0xFF;
-    return 1;
-  } else if (uc == 0xFFFE) {
-    dst[0] = 0xFE;
-    return 1;
+  // Note: we allow encoding 0xd800-0xdfff here, so as not to change
+  // the API, however, these are actually invalid in UTF-8
```
stevengj: In that case, do we still need `utf8proc_unsafe_encode_char`?

tkelman:
> checking for regressions on each individual test?

Not sure exactly what you mean, but the main thing a nicer test framework like clar (and there are even fancier ones out there, I'm sure) would do is run all the tests and give you a summary of failures at the end.

ScottPJones:
I just started looking at clar, but (in my really quick 50,000 feet flying look), it didn't seem to have built in tracking of previous results... but 1) I'm jetlagged, 2) need to look further...
The way the InterSystems test suite worked (IIRC) is that it would tag a potential release point (for the nightly build, for example), using Perforce, and run the tests, saving the results of all the tests...
(it also had files that had the correct results).  It would then compare (and it had to check if tests were added or deleted) each item, to see if the result changed...  if it changed, and didn't match the *correct* result, that was deemed a regression, and that build would be considered failed.
You'd get a nice report with how many new tests passed, how many new tests failed (along with more information to help debug each individual one), how many old tests passed, and how many failed (but did not change the result), and how many changed behavior and failed...  you could also run the test suite locally... or subsections of it...

ScottPJones:
Oh, I forgot to say, it compared with the last known good build along that branch... then if there are no regressions detected, the build is marked OK, and it's results (if different), are saved and become the new point for future comparisons in the branch.

ScottPJones:
Thanks!

PR #36, opened 2015-05-21
Respect user choice of MAKE, AR and INSTALL in Makefile

cookiemon:
Made the Makefile not overwrite the user's MAKE, AR and INSTALL variables.
Especially MAKE is important, since the file is not compatible with berkley make (because of the ifeq) and on FreeBSD GNU make is called gmake.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,9 +1,9 @@
 # libutf8proc Makefile
 
 # programs
-MAKE=make
-AR=ar
-INSTALL=install
+MAKE?=make
+AR?=ar
+INSTALL?=install
 
 # compiler settings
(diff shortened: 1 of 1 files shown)


stevengj:
I'm confused.  If you run `make MAKE=gmake` etcetera, then it [already overrides](http://www.gnu.org/software/make/manual/make.html#Overriding) `MAKE=make` lines in the Makefile, no?

I thought `=?` was only needed for makefiles overriding other makefile variables (e.g. for including one Makefile in another).

cookiemon:
Oh sorry, my mistake. Just forget this pull request ever existed.

kmsquire:
@cookiemon, no worries!  You should be able to close it yourself at the bottom (look for the "Close and comment" button).

PR #37, opened 2015-05-25
Enhance build process

jlec:
* Allow optimization flgas in CFLAGS to be overwritten
* Use Uppercase CC and CFLAGS
* Create all soname symlinks

Signed-off-by: Justin Lecher <jlec@gentoo.org>

### DIFF
--- a/data/data_generator.rb
+++ b/data/data_generator.rb
@@ -268,7 +268,7 @@ for code in 0...0x110000
   end
 end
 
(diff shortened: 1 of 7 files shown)


tkelman:
There are 5 other occurrences of `$(cc)` in this makefile that would also need to be changed here.

jlec:
Some rebase problems made this PR close, but #38 is the new one.

PR #38, opened 2015-05-29
Enhance build process

jlec:
* Allow optimization flgas in CFLAGS to be overwritten
* Use Uppercase CC and CFLAGS
* Create all soname symlinks

Signed-off-by: Justin Lecher <jlec@gentoo.org>

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -2,12 +2,13 @@
 
 # programs
 MAKE=make
-AR=ar
+AR?=ar
+CC?=gcc
 INSTALL=install
 
 # compiler settings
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -2,12 +2,13 @@
 
 # programs
 MAKE=make
-AR=ar
+AR?=ar
```
stevengj: Again, why is this needed?   `AR=ar` can be overridden by `make AR=....` already.

jlec: Yes, in case your environment has everything set, you just need to run ```make```. Especially when doing crosscompilation it is important not to call e.g. ```ar``` but the target platform specific version.

Issue #50, opened 2015-06-26
tag 1.3 release


stevengj:
It seems like we should be ready for this.  To do:

- [x] merge #51 in some form
- [x] update NEWS and README files as needed
- [x] push "v1.3" tag
- [x] update the web site (`gh-pages` branch): version numbers, release notes (copied from NEWS), documentation (run `doxygen` to generate).

Is there anything else or should I go ahead and do this?

tkelman:
Just to be on the safe side, could make sense to run some of Julia's tests locally against the master version of utf8proc?

jiahao:
Dare we ask if this works on Windoze...?

stevengj:
I will run Julia's tests (on MacOS).

jiahao:
Passes on Julia.MIT.edu (Ubuntu 15.04 64bit)

stevengj:
According to the [unifont maintainer](https://savannah.gnu.org/bugs/index.php?45395), they will release a new version with Unicode 8 support this Sunday, so I think we should hold off a couple of days so that we can update the charwidth tables.

stevengj:
Julia passes on MacOS.

tkelman:
win32 works fine too. win64 builds and is running tests now. I expect it should be okay, the changes since the previous dev version aren't that dramatic.

stevengj:
Done.

PR #51, opened 2015-06-29
Update character widths from Unifont 8.0.01

jiahao:
Ref #50


stevengj:
The Ruby oddness strikes again.  A difference in the `utf8proc_stage2table`, it seems like, but looking at the Ruby script I'm not sure why this would be non-deterministic.

@StefanKarpinski, you used to be a Ruby hacker, right?  Any idea why the data_generator.rb script would be non-deterministic?

stevengj:
Note that e.g. u+06de now has a charwidth of 2 with this patch.  This seems odd to me: on my terminal, `۞` seems to have a charwidth of 1.

jiahao:
Unifont 8.0.01 does define U+06DE to have fullwidth (1024 units); 7.0.06 gave it a negative RBearing of -1024 units so that it had width 0.

screen shot 2015-06-30 at 4 15 48 pm

StefanKarpinski:
Hashes are unordered in Ruby 1.8.x (but not, I believe in 1.9.x). Perhaps that could affect things?

stevengj:
@StefanKarpinski, I figured that the order of a hash would be undefined, but I'm not seeing where the script would depend on this.

stevengj:
Has anyone been able to reproduce this failure locally?  I tried using the same Ruby version as on Travis (ruby 1.9.3p551, from #48), but it still passes.

stevengj:
My inclination is to merge this and to open a separate issue for the data-generation Travis test, since that was a problem prior to this patch.

ScottPJones:
How difficult is this Ruby code to understand?  We've got a much better language right here!

stevengj:
I don't have the time to rewrite it from scratch right now, do you?

ScottPJones:
It is really that complicated? I'm just asking, not trying to pressure you (or anybody) into rewriting it.
I was just wondering if the effort of finding that bug would be more than getting something in Julia that would save time in the future (and not require a person with Ruby expertise).

StefanKarpinski:
The Ruby code is pretty straightforward, but it's also unclear why it would be nondeterministic. I don't think completely rewriting it is the answer.

stevengj:
Merged since the Ruby problem on Travis is not a regression in this patch.

PR #43, opened 2015-06-07
Always pass -fPIC in CFLAGS

nalimilan:
Fixes #42.

PR #53, opened 2015-09-12
always pass CFLAGS from environment

vtjnash:
ref https://github.com/JuliaLang/utf8proc/pull/43#issuecomment-109862458

(need to run CI on this since I edited from GitHub GUI)

### DIFF
--- a/bench/Makefile
+++ b/bench/Makefile
@@ -1,14 +1,14 @@
 CURL=curl
 
 CC = cc
-CFLAGS = -O2 -std=c99 -pedantic -Wall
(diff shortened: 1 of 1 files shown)


stevengj:
Probably it should (a) mimic the top-level Makefile and have a C99FLAG etc and (b) the `make -C bench` in top-level makefile should be pass these flags.

vtjnash:
Replaced by https://github.com/JuliaStrings/utf8proc/pull/43

vtjnash:
Sorry, wrong link. Replaced by https://github.com/JuliaStrings/utf8proc/pull/141

PR #55, opened 2015-10-29
Do not export internal unsafe_encode_char()

petercolberg:
This commit removes a spurious external symbol for the internal function `unsafe_encode_char()`, which I noticed while preparing a package of the utf8proc library for Debian.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -10,7 +10,8 @@ INSTALL=install
 CFLAGS ?= -O2
 PICFLAG = -fPIC
 C99FLAG = -std=c99
-UCFLAGS = $(CFLAGS) $(PICFLAG) $(C99FLAG) -DUTF8PROC_EXPORTS
+WCFLAGS = -Wall -Wmissing-prototypes -pedantic
+UCFLAGS = $(CFLAGS) $(PICFLAG) $(C99FLAG) $(WCFLAGS) -DUTF8PROC_EXPORTS
 
 # shared-library version MAJOR.MINOR.PATCH ... this may be *different*
 # from the utf8proc version number because it indicates ABI compatibility,
--- a/test/tests.h
+++ b/test/tests.h
@@ -33,7 +33,6 @@ size_t skipspaces(const char *buf, size_t i)
    separated by whitespace, and terminated by any character not in
    [0-9a-fA-F] or whitespace, then stores the corresponding utf8 string
    in dest, returning the number of bytes read from buf */
-utf8proc_ssize_t unsafe_encode_char(utf8proc_int32_t uc, utf8proc_uint8_t *dst);
 size_t encode(char *dest, const char *buf)
 {
      size_t i = 0, j, d = 0;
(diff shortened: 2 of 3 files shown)


petercolberg:
Strange, the tests with `make check` are passing, but the `make data` target is failing.

petercolberg:
I added the compiler flag `-Wmissing-prototypes` that warns in case of exported internal functions:
```
cc -O2 -Wall -Wmissing-prototypes -fPIC -std=c99 -pedantic -DUTF8PROC_EXPORTS -c -o utf8proc.o utf8proc.c
utf8proc.c:191:18: warning: no previous prototype for ‘unsafe_encode_char’ [-Wmissing-prototypes]
 utf8proc_ssize_t unsafe_encode_char(utf8proc_int32_t uc, utf8proc_uint8_t *dst) {
                  ^
```

stevengj:
Yes, we've had some persistent problems with the `make data` check because the ruby script seems to be producing non-deterministic tables, probably due to differences in hashing algorithms in different ruby versions or on different systems; we haven't tracked it down yet.

stevengj:
The appveyor test seems to have bitrotted as well, grrr....

petercolberg:
Just to make sure that my change does not introduce a silent regression to the tests: Was there any particular reason why the tests were using `unsafe_encode_char()` instead of `utf8proc_encode_char()`?

Makefile
```diff
@@ -10,7 +10,8 @@ INSTALL=install
 CFLAGS ?= -O2
 PICFLAG = -fPIC
 C99FLAG = -std=c99
-UCFLAGS = $(CFLAGS) $(PICFLAG) $(C99FLAG) -DUTF8PROC_EXPORTS
+WCFLAGS = -Wall -Wmissing-prototypes -pedantic
```
tkelman: Is `-Wmissing-prototypes` not included in `-Wall`? If so, why not? MSVC understands `-Wall` but not `-Wmissing-prototypes`, and currently my hacky method of building Julia with MSVC goes through makefiles rather than cmake for some deps.

stevengj: It's not included in `-Wall` [for gcc](https://gcc.gnu.org/onlinedocs/gcc-6.1.0/gcc/Warning-Options.html#Warning-Options).

stevengj: If it causes problems with MSVC, I would just drop this warning.

petercolberg: Better suggestion: move `-Wmissing-prototypes` to `CFLAGS` in `.travis.yml`.

The flag serves a purpose, to warn when internal functions are accidentally exported.

tkelman: I like that plan, as long as people will skim the travis logs once in a while

petercolberg: See `.travis.yml`, the flag `-Werror` ensures that Travis will fail on warnings.

The 2.0 release and subsequent #77 proved that warnings will be ignored with certainty unless fatal. :wink:

PR #57, opened 2015-10-30
Fix MingW build test

petercolberg:
Fix MingW build test

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -22,16 +22,16 @@ build_script:
   - cd ..\msvc_shared
   - cmake .. -DBUILD_SHARED_LIBS=ON
   - cmake --build .
+  - set PATH=C:\MinGW\bin;%PATH%
   - C:\MinGW\msys\1.0\bin\sh --login -c "
-      echo 'C:\MinGW\ /MinGW' > /etc/fstab;
-      cd /c/projects/utf8proc;
-      mkdir mingw_static;
-      cd mingw_static;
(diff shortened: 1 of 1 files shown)


petercolberg:
Is it possible to lower the appveyer build timeout to say 5 minutes?

The current build hangs, and will only timeout after 30 or 40 minutes.

aviks:
@petercolberg I've added the [WIP] tag to the title to indicate that this is experimental.  Please take it out once this PR is ready.

tkelman:
The problem is that `cd /c/projects/utf8proc` is failing in msys, not the choice of generator. mingw make is a crippled version that is enough to build utf8proc but not much else.

petercolberg:
When the line that creates /etc/fstab is removed, then `cd /c/projects/utf8proc` succeeds.

The build freezes as soon as `cmake … -G'MSYS Makefiles'` is executed.

petercolberg:
I could finally trigger the following error message in the interactive session:
```
cc1.exe: The program can't start because libgmp-10.dll is missing from your computer.
Try reinstalling the program to fix this problem.
```
This is why cmake hangs when using the MSYS Makefiles generator.

tkelman:
That sounds like a broken toolchain installation, cc @feodorfitsner. So changing the generator must be either changing which mingw compiler is getting run, or modifying the path so it's able to find libgmp.

petercolberg:
It’s already fixed. Please merge after the Travis build succeeds.

stevengj:
Looks great, thanks for following up on this!

PR #58, opened 2015-10-30
Fix build warnings

petercolberg:
This commit series fixes several build warnings, and ensures that `make clean` reverts the build tree to a pristine state. This is the last pull request in an effort to bring utf8proc into shape for distribution packaging, specifically for Debian.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -44,6 +44,9 @@ all: libutf8proc.a libutf8proc.$(SHLIB_EXT)
 
 clean:
 	rm -f utf8proc.o libutf8proc.a libutf8proc.$(SHLIB_VERS_EXT) libutf8proc.$(SHLIB_EXT) test/normtest test/graphemetest test/printproperty test/charwidth test/valid test/iterate
+ifneq ($(OS),Darwin)
+	rm -f libutf8proc.so.$(MAJOR)
+endif
 	$(MAKE) -C bench clean
 	$(MAKE) -C data clean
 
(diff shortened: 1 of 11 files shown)


nalimilan:
Thanks! This looks good to me, apart from the few comments I have added inline. Also, it's funny that the test you added to check that `make clean` does its job fails on Linux and Mac...

petercolberg:
Ready for review.

Makefile
```diff
@@ -84,12 +93,17 @@ install: libutf8proc.a libutf8proc.$(SHLIB_EXT) libutf8proc.$(SHLIB_VERS_EXT)
...
-	ln -f -s $(libdir)/libutf8proc.$(SHLIB_VERS_EXT) $(DESTDIR)$(libdir)/libutf8proc.$(SHLIB_EXT)
+	cd $(DESTDIR)$(libdir) && ln -f -s libutf8proc.$(SHLIB_VERS_EXT) libutf8proc.$(SHLIB_EXT)
 ifneq ($(OS),Darwin)
-	ln -f -s $(libdir)/libutf8proc.$(SHLIB_VERS_EXT) $(DESTDIR)$(libdir)/libutf8proc.so.$(MAJOR)
-	ln -f -s $(libdir)/libutf8proc.$(SHLIB_VERS_EXT) $(DESTDIR)$(libdir)/libutf8proc.so.$(MAJOR).$(MINOR)
+	cd $(DESTDIR)$(libdir) && ln -f -s libutf8proc.$(SHLIB_VERS_EXT) libutf8proc.so.$(MAJOR)
 endif
 
+MANIFEST.new:
+	rm -rf INSTALL
```
stevengj: Probably call it `tmp` or similar rather than `INSTALL`.

(In GNU packages, `INSTALL` is a file with installation instructions.)

petercolberg: I have renamed the directory to `tmp`.

test/tests.h
```diff
@@ -1,5 +1,6 @@
 /* Common functions and includes for our test programs. */
 
+#define _XOPEN_SOURCE 700
```
stevengj: Would be nice to add a comment for why this is needed.

petercolberg: Done.

stevengj:
Thanks!

PR #56, opened 2015-10-30
Fix Travis build and improve build reproducibility

petercolberg:
The first commit fixes the Travis builds, which are currently failing. I could also reproduce the updated data file on Debian unstable using Ruby 2.1.5p273. The subsequent commits make the process of generating the data file more robust, which is desirable for distribution packagers that wish to reproduce the entire build process.

PR #59, opened 2015-11-01
Version 1.3.1

petercolberg:
It would be nice if you could tie a ribbon around #55, #56, #57, #58 by tagging a patch release. A tagged version makes it much easier to package utf8proc for Debian, which is a prerequisite for getting Julia 0.4.0 into Debian. I updated NEWS with all user-visible changes.


stevengj:
LGTM.

stevengj:
Also tagged and updated the web page.

PR #62, opened 2015-12-09
Reduce the size of the binary.

michaelnmmeyer:
A few changes aimed at shrinking the compiled library. Saves 226 kb / 716 kb.
A downside is that this breaks binary compatibility.


stevengj:
Sounds reasonable.   The current length of `utf8proc_sequences` seems to be 12461, so we are pretty far from 65535 (and we can always change the size if we have to in the future).

Because this breaks ABI compatibility, but preserves the documented API, you should change the major soversion at https://github.com/JuliaLang/utf8proc/blob/6b510deff274dc79b4e324abef5fbb45a8162861/Makefile#L22-L24 (i.e. major=2, minor=patch=0), and correspondingly in cmakelists and manifest as mentioned in that file.  (Leave the API version unchanged for now; we'll bump the minor version in the next release.)

michaelnmmeyer:
Okay, version bumped.

ScottPJones:
This is very nice!  When I worked on fixing some bugs in `utf8proc` months ago, I'd commented on the tables not being efficient, but didn't have time to do anything about it.  :+1:

PR #65, opened 2016-01-22
Don't leak compiler settings to other CMake targets

per-gron:
It is a global property, which is not great when importing utf8proc in other CMake projects.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -16,6 +16,8 @@ add_library (utf8proc
   utf8proc.h
 )
 
+target_include_directories(utf8proc PUBLIC ../utf8proc)
+
 set_target_properties(
   utf8proc PROPERTIES
   COMPILE_DEFINITIONS "UTF8PROC_EXPORTS"
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -11,13 +11,14 @@ set(SO_MAJOR 2)
 set(SO_MINOR 0)
 set(SO_PATCH 0)
 
-add_definitions (
-  -DUTF8PROC_EXPORTS
-)
-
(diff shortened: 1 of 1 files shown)


tkelman:
What's the minimum cmake version for `target_include_directories` ?

pereckerdal:
@tkelman it requires CMake 2.8.11 it seems. Is that a problem? http://www.kitware.com/blog/home/post/492

nalimilan:
RHEL6 has 2.8.12.2, and RHEL5 2.8.11.2 via EPEL5's make28 package. So this change looks fine to me.

tkelman:
Ubuntu 12.04 only has cmake 2.8.7, and that's what travis ci uses by default. Since cmake is optional here I think it's okay to add the toolchain ppa and a newer cmake package via the travis apt addon.

nalimilan:
I wouldn't have thought Ubuntu 12.04 would lag behind RHEL5... Though a PPA provides 2.8.11.2: https://launchpad.net/~kalakris/+archive/ubuntu/cmake

tkelman:
the default cmake on rhel5 is 2.6.

if travis runs on container infrastructure here then you can't use sudo and add arbitrary ppa's, only those on the whitelist.

pereckerdal:
It sounds to me like maybe this won't work with your current CI infrastructure? Perhaps it's easiest to simply close this and maybe revisit once Travis has upgraded its Ubuntu version?

tkelman:
It can be made to work. Just need to manually install a newer cmake apt package via something like
```
addons:
 apt:
   sources:
   - kalakris-cmake
   packages:
   - cmake
```
in `.travis.yml`

PR #66, opened 2016-02-04
Fix overrun

michaelnmmeyer:
It went undetected so far because the corresponding test program was reusing the same initialized
memory buffer several times.


stevengj:
Yikes, good catch.

Issue #67, opened 2016-05-25
Minimizing tables for (de)composition


benibela:
So I was using a Pascal port of utf8proc to convert utf8 strings between the composed and decomposed form.

Then I noticed the Unicode tables are 600kb large (also due to the port not doing bitpacking), which is not good.

So I dropped every field from the struct that was not used for compositing/decomposition, who needs title case or folding anyways? 
Then the tables were only slightly above 300kb large. Much better.

But looking closer at the tables, half the sequence array consists of -1. Those values are not needed for anything. So I removed all the -1, moving the element length to an unused byte in the property array, saving 20kb and bringing the tables to under 300kb. 

Then the combination array. It is even worse. Almost all of its entries are -1. Someone lost their -1 collection. 
A sparse [1st index][2nd index] array, all 2nd arrays joined together, indexed as [1st index*size + 2nd index]. You could cut the -1 prefix/suffix of each 2nd array, making them variable sized and index it as [startindex[1st index] + 2nd index]. But since I care more about space than performance, I replaced the 2nd array with a list of (2nd index, value) pairs. Now you need a linear search starting at startindex[1st index] to find the 2nd index and then get the value, but it is much smaller. 
In fact it reduces this table size to 8kb, saving over 90kb.

Now the property struct contains a 1st and 2nd comb index. Yet there are no characters that have both. I replaced it by a single field comb index, saving 5 kb. Pointless

The stage2 array has a curious property. It is full of almost alternating 0 and 2 elements.
Why? 0 is the default property struct used for missing (?) characters, 2 is the property struct for characters without special properties. Both structs are the same.
Dropping struct 2, turned these stage2 blocks into blocks of zeros, which were then automatically merged by the data generator, saving 20kb.

Back to the sequence array. It is a list of 32-bit numbers, but those numbers are small, wasting a lot of zero bits.
It is like an UTF-32 string. I changed it to UTF-16, which saves 10kb.

The decomposition function does not actually use the decomp_type, only checks if it is zero or not.
I do not need that type. Replacing it with a bool, automatically drops 800 property structs, as they become duplicates.

Now, what exactly is the property array? After removing all that stuff, it is actually a map of codepoint -> decompositon index.
Some structs have a composition index, but only like 10% of them. No point in storing that index for the properties that do not have it. So I replaced the property array with a DWORD-array, bit encoding the properties. Fields with comp index take 2 entries, most only need 1 entry.
This saves 20kb.

[...]

PR #68, opened 2016-05-31
Smaller tables

benibela:
See #67


ivarne:
Seems like travis is failing because of an expired cert
```
# use JuliaLang caching (https://github.com/staticfloat/cache.julialang.org)
# so that Travis builds do not depend on anyone's flaky servers but our own
    - URLCACHE=https://cache.e.ip.saba.us/
```

cc: @staticfloat

ivarne:
The travis failiure will probably be fixed by https://github.com/JuliaLang/utf8proc/pull/69

tkelman:
It might need a rebase and force push though. Changes to `.travis.yml` might not propagate automatically by just restarting the build.

stevengj:
Needs a version bump.  Since this is a technically a backwards incompatible change, we'll need to change the major version? 

Maybe we should take the opportunity to export other accessor functions, like `utf8proc_casefold(codepoint)`, `utf8proc_bidi(codepoint)`, ...?

We could unexport the `utf8proc_property_struct` entirely, replacing it by accessor functions.  Theoretically, this would make it less efficient to extract multiple properties of a codepoint at once, but I'm skeptical that this is common in practice.

stevengj:
is this ready to merge?

benibela:
Unless you want to increase the version number first for people using the property struct directly.

And Unicode 9 was just released. Might be worth checking if it still fits in the tables. Some are almost full.

stevengj:
Needs a rebase and update now that Unicode 9 support has been merged.

utf8proc.h
```diff
@@ -550,6 +549,13 @@ UTF8PROC_DLLEXPORT utf8proc_int32_t utf8proc_tolower(utf8proc_int32_t c);
 UTF8PROC_DLLEXPORT utf8proc_int32_t utf8proc_toupper(utf8proc_int32_t c);
 
 /**
+ * Given a codepoint `c`, return the codepoint of the corresponding
+ * title-case character, if any; otherwise (if there is no title-case
+ * variant, or if `c` is not a valid codepoint) return `c`.
+ */
+UTF8PROC_DLLEXPORT utf8proc_int32_t utf8proc_totitle(utf8proc_int32_t c);
```
stevengj: Should this be equivalent to `toupper` if there is no titlecase?

stevengj:
Awesome, thanks for updating this.  Did you have any problems with the table size?

stevengj:
What's the overall reduction in (compiled) table size from this patch?

benibela:
200 k (42%)

data/data_generator.rb
```diff
@@ -115,22 +115,52 @@ def str2c(string, prefix)
...
+def cpary2c(array)
+  return "UINT16_MAX" if array.nil? || array.length == 0
+  lencode = array.length - 1 #no sequence has len 0, so we encode len 1 as 0, len 2 as 1, ... 
+  array = cpary2utf16encoded(array)
+  if lencode >= 7 #we have only 3 bits for the length (which is already cutting it close. might need to change it to 2 bits in future Unicode versions)
+    array = [lencode] + array 
+    lencode = 7
+  end  
+  idx = pushary(array) 
+  raise "Array index out of bound" if idx > 0x1FFF
```
stevengj, line 147: This exception triggered in Unicode 14.   @benibela, what do you suggest?

benibela: @stevengj one bit needs to be taken from lencode (7 -> 3) and given to idx (0x1FFF -> 0x3FFF, 13 -> 14)

and the same in seqindex_write_char_decomposed

stevengj: Could you make a PR?

PR #70, opened 2016-06-24
Unicode 9 updates

Keno:
I believe the only substantiative changes to the code required are the new rules in TR29, but please do check the Unicode change log (http://www.unicode.org/versions/Unicode9.0.0/) for other changes that may affect this library.


utf8proc.c
```diff
@@ -233,33 +233,56 @@ UTF8PROC_DLLEXPORT const utf8proc_property_t *utf8proc_get_property(utf8proc_int
...
 }
 
-/* return whether there is a grapheme break between codepoints c1 and c2 */
+/* return whether there is a grapheme break between codepoints c1 and c2.
+
+  CAVEATS:
+   Please note that evaluation of GB10 (grapheme breaks between emoji zwj sequences)
+   and GB 12/13 (regional indicator code points) require knowledge of previous characters
+   and are thus not handled by this function. This may result in an incorrect break before
+   and E_Modifier class codepoint and an incorrectly missing break between two
```
tkelman: an E_modifier?

tkelman:
This almost certainly needs version number bumps

utf8proc.c
```diff
@@ -389,7 +412,21 @@ UTF8PROC_DLLEXPORT utf8proc_ssize_t utf8proc_decompose_char(utf8proc_int32_t uc,
...
+    if (*last_boundclass == tbc && tbc == UTF8PROC_BOUNDCLASS_REGIONAL_INDICATOR)
+      *last_boundclass = UTF8PROC_BOUNDCLASS_OTHER;
+    // Special support for GB10. Fold any EXTEND codepoints into the previous
+    // boundclass if we're dealing with an emoji base boundclass.
+    else if ((*last_boundclass == UTF8PROC_BOUNDCLASS_E_BASE      ||
+              *last_boundclass == UTF8PROC_BOUNDCLASS_E_BASE_GAZ) &&
+             tbc == UTF8PROC_BOUNDCLASS_EXTEND)
+      *last_boundclass = UTF8PROC_BOUNDCLASS_E_BASE;
+    else
+      *last_boundclass = tbc;
```
stevengj: Would be nice to export this logic somehow so that we can use it in the Julia grapheme iterator.

Keno: Mhm, I believe just adding an extra out parameter to give you the override class should be fine.

Keno:
Ok, I've updated the API to expose the state override. Since that's an ABI incompatible change, I've bumped the MAJOR version accordingly. Please review.

Keno:
We might want to do this at the same time as #68 (which may also require a rebase for this). cc @benibela

Makefile
```diff
@@ -19,9 +19,9 @@ UCFLAGS = $(CFLAGS) $(PICFLAG) $(C99FLAG) $(WCFLAGS) -DUTF8PROC_EXPORTS
 # not API compatibility: MAJOR should be incremented whenever *binary*
 # compatibility is broken, even if the API is backward-compatible
 # Be sure to also update these in MANIFEST and CMakeLists.txt!
```
stevengj: Did you forget to update CMakeLists.txt too?

Keno: Yes

stevengj: Note that you also need to update `utf8proc.h` for the API version change.

stevengj:
Maybe better to do this first, then update #68?

utf8proc.h
```diff
@@ -520,8 +520,19 @@ UTF8PROC_DLLEXPORT utf8proc_ssize_t utf8proc_reencode(utf8proc_int32_t *buffer,
...
+ *
+ * @param state Beginning with Version 29 (Unicode 9.0.0), this algorithm requires
+ *              state to break graphemes. This state can be passed in as a pointer
+ *              in the `state` argument and should initially be set to 0. If the
+ *              state is not passed in (i.e. a null pointer is passed), UAX#29 rules
+ *              GB10/12/13 which require this state will not be applied, essentially
+ *              matching the rules in Unicode 8.0.0.
+ *
+ * @warning If the state parameter is used, `utf8proc_grapheme_break` must be called
+ *          IN ORDER on ALL potentital breaks in a string.
```
stevengj: potential

utf8proc.c
```diff
@@ -273,19 +282,38 @@ static utf8proc_bool grapheme_break(int lbc, int tbc) {
...
 }
 
-/* return whether there is a grapheme break between codepoints c1 and c2.
+static utf8proc_bool grapheme_break_extended(int lbc, int tbc, utf8proc_int32_t *state)
+{
+  int lbc_override = lbc;
+  if (state && *state != UTF8PROC_BOUNDCLASS_START)
+    lbc_override = *state;
+  utf8proc_bool break_permitted = grapheme_break_simple(lbc, tbc);
+  if (state) {
```
stevengj: Is it really necessary to have general state like this?  Wouldn't it be enough just to know the previous codepoint before `lbc`?

Keno: No, there may be an infinite number of EXTEND characters between E_Base and E_Modifier.

Keno:
Do we need to hold off on doing the version bump if we want to do #68 right after?

tkelman:
I think we can do a version bump for this (definitely needed), and if we want to merge further big changes we don't need to change the version again until we do the next release.

Keno:
Ok.

Keno:
Good to merge?

benibela:
There is a warning about an unused variable: lbc_override

Keno:
Oops, thanks.

Keno:
I also realized I made this static. Will fix.

utf8proc.c
```diff
@@ -233,36 +233,87 @@ UTF8PROC_DLLEXPORT const utf8proc_property_t *utf8proc_get_property(utf8proc_int
...
 
-/* return whether there is a grapheme break between codepoints c1 and c2 */
-UTF8PROC_DLLEXPORT utf8proc_bool utf8proc_grapheme_break(utf8proc_int32_t c1, utf8proc_int32_t c2) {
-  return grapheme_break(utf8proc_get_property(c1)->boundclass,
-                        utf8proc_get_property(c2)->boundclass);
+static utf8proc_bool grapheme_break_extended(int lbc, int tbc, utf8proc_int32_t *state)
+{
+  int lbc_override = lbc;
+  if (state && *state != UTF8PROC_BOUNDCLASS_START)
+    lbc_override = *state;
```
stevengj: @Keno, I just noticed that `lbc_override` seems to never be used.  Did you mean to pass `lbc_override` to `grapheme_break_simple`?

Keno: It's been a while, but yes, that looks right, esp by looking at the first commit in this PR.

Keno: We already seem to have found that bug though: https://github.com/JuliaLang/utf8proc/blob/master/utf8proc.c#L290

Keno: https://github.com/JuliaLang/utf8proc/pull/77

Issue #72, opened 2016-07-11
Tag release with Unicode 9 support


petercolberg:
utf8proc fails to build on Debian due to not supporting the current Unicode data ([#829236](https://bugs.debian.org/829236)). As the bug is release critical, utf8proc and Julia are scheduled to be removed from Debian testing on 2016-08-14.

Could you tag a new release that includes support for Unicode 9 (#70)?

stevengj:
Yes, we should definitely do this soon.  (We need to bump the GNU Unifont version used to compute font metrics first.)

In the future, it would be good to download test data only for the version of Unicode that we actually support, rather than the latest version.  Is there an archive URL for the test data from earlier Unicode versions?

stevengj:
One concern is that older versions of Julia will not currently build with this version, due to changes in the `grapheme_break` API.

stevengj:
Okay, I've updated the `NEWS`.   I noticed that we accidentally bumped the ABI version twice, once in #62 and once in #70, even though there was no intervening release, so I've changed the ABI version back to 2.0.0.

@Keno, anything else I'm forgetting before we can release?

stevengj:
@tkelman, can you update the README with an AppVeyor status badge?     Since it seems to be [set up under your account](https://ci.appveyor.com/project/tkelman/utf8proc), I'm not sure how to get the status badge URL.

Keno:
Can't think of anything you haven't covered :shipit:.

stevengj:
Okay, v2.0 is tagged and released.

petercolberg:
@stevengj I still get this warning:
```
 cc -O2 -fPIC -std=c99 -Wall -Wmissing-prototypes -pedantic -DUTF8PROC_EXPORTS -c -o utf8proc.o utf8proc.c
utf8proc.c: In function ‘grapheme_break_extended’:
utf8proc.c:287:7: warning: variable ‘lbc_override’ set but not used [-Wunused-but-set-variable]
   int lbc_override = lbc;
       ^
```

stevengj:
@Keno, what's up with that warning?  Looks like some dead code?

Keno:
Damn, yes we forgot to fix that.

stevengj:
If it is just a harmless warning, I don't think it requires a new release, but please patch it.

Keno:
Unfortunately it's not harmless for people who use the stateful api directly, but it does happen work for the internal code because `lbc == *state`.

[...]

PR #75, opened 2016-07-12
update to Unifont 9 (for Unicode 9 charwidths)

stevengj:
Since we bumped to Unicode 9 (in #70), we should also update to the latest version of GNU Unifont when computing character widths.

This is needed to tag a new release (#72).


stevengj:
@jiahao, what's going on with AppVeyor?

stevengj:
Okay, seems like AppVeyor problems were just a temporary glitch on their side; I saw the same failure on multiple projects.

tkelman:
doesn't unicode 9 address the char width issue we needed this unifont stuff to address?

Keno:
Most of it.

Issue #24, opened 2015-03-06
shared-library versioning


stevengj:
We should really include a version number in the shared library (this is different from the utf8proc source version: the source version indicates API compatibility, while the library version encodes ABI compatibility).     I have used [libtool for this](http://www.gnu.org/software/libtool/manual/html_node/Updating-version-info.html) in the past, but the whole autotools stack is probably not worth the trouble here (although we could use just the `libtool` script, which could be bundled by `make dist`).

stevengj:
See also the [discussion of a soname in the Fedora package](https://bugzilla.redhat.com/show_bug.cgi?id=1058019).   cc: @nalimilan

nalimilan:
Great! Note that by choosing 1.2.0, you signal an ABI break as compared with the 0.1 SOVERSION I have used in the Fedora package. Is that intended?

stevengj:
Yes, that is intended.   The ABI has changed incompatibly because a `struct` was changed by #20 (although the API is backwards compatible).

Of course, since this is the first release with an soname, I could have chosen `1.0.0` or even `0.0.0` (though Darwin doesn't like zero versions), but I figured it was less confusing to start with a version that matches the API version (even if they subsequently diverge).

nalimilan:
OK, perfect. Indeed, following the API version when possible is a good idea. Ideally ABI and API will only be broken at the same time in the future, and both versions incremented together.

PR #76, opened 2016-07-12
Walk back ABI breaking changes

Keno:


CMakeLists.txt
```diff
@@ -7,8 +7,8 @@ disallow_intree_builds()
 project (utf8proc C)
 
 # Be sure to also update these in Makefile!
-set(SO_MAJOR 3)
-set(SO_MINOR 0)
+set(SO_MAJOR 2)
```
stevengj: Hasn't the ABI technically changed because of #68?

utf8proc.h
```diff
@@ -530,9 +530,16 @@ UTF8PROC_DLLEXPORT utf8proc_ssize_t utf8proc_reencode(utf8proc_int32_t *buffer,
  * @warning If the state parameter is used, `utf8proc_grapheme_break` must be called
```
stevengj: fix the docstring

utf8proc.h
```diff
@@ -530,9 +530,16 @@ UTF8PROC_DLLEXPORT utf8proc_ssize_t utf8proc_reencode(utf8proc_int32_t *buffer,
...
  *          IN ORDER on ALL potential breaks in a string.
  */
-UTF8PROC_DLLEXPORT utf8proc_bool utf8proc_grapheme_break(
+UTF8PROC_DLLEXPORT utf8proc_bool utf8proc_grapheme_break_stateful(
     utf8proc_int32_t codepoint1, utf8proc_int32_t codepoint2, utf8proc_int32_t *state);
 
+/**
+ * Same as utf8proc_grapheme_break_stateful, except without support for the Unicode 9
```
stevengj: use `@ref utf8proc_grapheme_break_stateful` to get a cross-reference here

CMakeLists.txt
Keno: I suppose is has.

stevengj: In particular, because of the changes to the `utf8proc_property_struct` data structure.   Or do we not consider those fields part of the public API?

Keno: ABI compatibility is the question "if we drop the new library in where the old one used to be will things keep working" and because of those changes, I think the answer is no.

stevengj: Note that we changed the SO version in #24 because of the `struct` changes in #20.

But maybe we should stop considering the `struct` to be part of the ABI?  Maybe we should stop exposing it in the header file.

Keno: Are all the properties in the struct accessible via methods?

Keno:
Ok, I've removed the ABI de-bump, which can be a separate decision.

CMakeLists.txt
stevengj: No.

PR #77, opened 2016-07-13
Fix incorrect use of `lbc` instead of `lbc_override`

Keno:


stevengj:
Thanks, I'll tag 2.0.1 shortly.

PR #79, opened 2016-07-16
Move -Wmissing-prototypes from Makefile to .travis.yml

tkelman:
since MSVC doesn't understand this flag, and the current
mechanism for building Julia with MSVC goes through the makefile

as suggested by @petercolberg at https://github.com/JuliaLang/utf8proc/pull/55#discussion_r71060437

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -19,4 +19,4 @@ script:
 env:
     # use JuliaLang caching (https://github.com/staticfloat/cache.julialang.org)
     # so that Travis builds do not depend on anyone's flaky servers but our own
-    - URLCACHE=https://cache.julialang.org/ CFLAGS="-O2 -Werror"
+    - URLCACHE=https://cache.julialang.org/ CFLAGS="-O2 -Werror -Wmissing-prototypes"
--- a/Makefile
+++ b/Makefile
@@ -11,7 +11,7 @@ FIND=find
 CFLAGS ?= -O2
 PICFLAG = -fPIC
 C99FLAG = -std=c99
-WCFLAGS = -Wall -Wmissing-prototypes -pedantic
+WCFLAGS = -Wall -pedantic
 UCFLAGS = $(CFLAGS) $(PICFLAG) $(C99FLAG) $(WCFLAGS) -DUTF8PROC_EXPORTS
 
 # shared-library version MAJOR.MINOR.PATCH ... this may be *different*


tkelman:
I shouldn't spend too much time getting to the bottom of this right now, but there might be a serious issue with utf8proc 2 and/or flisp or julia's use of it with msvc, even 2013 which has at least been able to get through Julia's bootstrap without all that many patches. If I try to use this branch of utf8proc with any version of Julia since the upgrade to utf8proc 2 was merged, I get
```
 cd /c/projects/julia/base && /c/projects/julia/usr/bin/julia.exe -C x86-64 --output-ji /c/projects/julia/usr/lib/julia/inference.ji --startup-file=no coreimg.jl
make[1]: *** [/c/projects/julia/usr/lib/julia/inference.ji] Error 253
```
or similar when trying to start bootstrap. Reverting https://github.com/JuliaLang/julia/commit/10359ec3159085f96ef95aeacae16c51b9543917 gets through bootstrap again, ref https://ci.appveyor.com/project/tkelman/julia/build/1.0.917 (failing to compile ccalltest.c at the end is expected). Anyone who's curious about this should be able to reproduce this by working off of https://github.com/tkelman/julia/commits/tk/avmsvc-clean and enabling appveyor on your personal fork of Julia.

stevengj:
Regarding the MSVC problem that's concerning, but I'm not sure what the problem could be.  On Linux, I ran the utf8proc tests through `valgrind` with no problems.

tkelman:
that won't exercise the msvc-only ifdefs, which might now be subtly wrong for the latest code - ref warnings noted in https://github.com/JuliaLang/utf8proc/issues/63#issuecomment-233121484

stevengj:
I just looked at how flisp is calling utf8proc, and it seems wonky and antiquated.  I'll put together a PR.

stevengj:
Though I don't see why it should crash.  I was worried about the low-level calls to `utf8proc_decompose` etcetera, but they seem fine.  The only change I want to make is that we should call `utf8proc_category` rather than directly accessing `utf8proc_property_t`, but I don't see why that should crash it.

tkelman:
Hm. Dunno, guess I can try getting it going locally and see if I can remember how to work windbg, but I've got higher-priority things to work on right now. Mostly checking if the back-burner "in case we need it" msvc support might actually catch a real bug once in a while.

stevengj:
Yeah, I totally agree.  Porting to multiple compilers is a great way to catch bugs.  (Unfortunately, sometimes you catch compiler bugs too.)

PR #80, opened 2016-07-26
use a different variable name for nested loop in bench.c

tkelman:
and declare it ahead of time to avoid "error: 'for' loop initial declarations are only allowed in C99 mode"

### DIFF
--- a/bench/bench.c
+++ b/bench/bench.c
@@ -7,7 +7,7 @@
 
 int main(int argc, char **argv)
 {
-	 int i;
+	 int i, j;
 	 int options = 0;
 	 
 	 for (i = 1; i < argc; ++i) {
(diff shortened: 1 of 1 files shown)


tkelman:
Any objection @stevengj? Would you mind if I tagged 2.0.2 after this? The changes have been pretty negligible since 2.0.1 but it'd be useful for https://github.com/JuliaLang/julia/pull/17637

stevengj:
Sure. File a PR that updates NEWS and the various version numbers before tagging.

tkelman:
Right, I knew I was forgetting something.

tkelman:
https://github.com/JuliaLang/utf8proc/pull/53 would probably also be able to address the build failure.

PR #81, opened 2016-07-26
NEWS and version numbers for 2.0.2

tkelman:


stevengj:
Also in utf8proc.h for the API version (which is in principle distinct from the ABI version in the Makefiles).

stevengj:
Okay, I tagged 2.0.2.

Will update the web page shortly.

tkelman:
thanks!

PR #85, opened 2016-09-26
ability create static library

nicolaichuk:
add msvc "/Wall"

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -13,18 +13,54 @@ set(SO_MAJOR 2)
 set(SO_MINOR 0)
 set(SO_PATCH 2)
 
-add_definitions (
-  -DUTF8PROC_EXPORTS
-)
+# configure build type
 
-if (NOT MSVC)
+set(_utf8proc_is_set_build_shared_libs TRUE)
+if(DEFINED BUILD_SHARED_LIBS)
+  set(_utf8proc_is_set_build_shared_libs ${BUILD_SHARED_LIBS})
+endif(DEFINED BUILD_SHARED_LIBS)
+
+set(UTF8PROC_BUILD_AS_SHARED ${_utf8proc_is_set_build_shared_libs} CACHE BOOL "If present and true, this will cause utf8proc to be built shared unless the utf8proc was explicitly added as a static library.")
+
+# set additional defined
+
+if(NOT DEFINED _utf8proc_additional_defined)
+  set(_utf8proc_additional_defined )
+endif(NOT DEFINED _utf8proc_additional_defined)
+
+if(UTF8PROC_BUILD_AS_SHARED)
+  list(APPEND _utf8proc_additional_defined 
+    "-DUTF8PROC_EXPORTS"
+  )
+  set(_utf8proc_build_type "SHARED")
+else(UTF8PROC_BUILD_AS_SHARED)
+  list(APPEND _utf8proc_additional_defined 
+    "-DUTF8PROC_BUILD_AS_LIB"
+  )
+  set(_utf8proc_build_type "STATIC")
+endif(UTF8PROC_BUILD_AS_SHARED)
+list(REMOVE_DUPLICATES _utf8proc_additional_defined)
+add_definitions("${_utf8proc_additional_defined}")
+
+# other config
+
+if(NOT MSVC)
   set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -O2 -std=c99 -pedantic -Wall")
-endif ()
+endif()
 
-add_library (utf8proc
+add_library (utf8proc ${_utf8proc_build_type}
   utf8proc.c
   utf8proc.h
 )
+
+if(MSVC)
+  get_target_property(_utf8proc_msvc_build_prop_val utf8proc COMPILE_FLAGS)
+  if(${_utf8proc_msvc_build_prop_val} STREQUAL "_utf8proc_msvc_build_prop_val-NOTFOUND")
+    set_target_properties(utf8proc PROPERTIES COMPILE_FLAGS "/Wall")
+  else(${_utf8proc_msvc_build_prop_val} STREQUAL "_utf8proc_msvc_build_prop_val-NOTFOUND")
+    set_target_properties(utf8proc PROPERTIES COMPILE_FLAGS "${_utf8proc_msvc_build_prop_val} /Wall")
+  endif(${_utf8proc_msvc_build_prop_val} STREQUAL "_utf8proc_msvc_build_prop_val-NOTFOUND")
+endif()
 
 set_target_properties (utf8proc PROPERTIES
   POSITION_INDEPENDENT_CODE ON
--- a/utf8proc.h
+++ b/utf8proc.h
@@ -113,16 +113,24 @@ typedef bool utf8proc_bool;
 #endif
 #include <limits.h>
 
-#ifdef _WIN32
-#  ifdef UTF8PROC_EXPORTS
-#    define UTF8PROC_DLLEXPORT __declspec(dllexport)
+#ifndef UTF8PROC_BUILD_AS_LIB
+#  ifdef _WIN32
+#    ifdef UTF8PROC_EXPORTS
+#      define UTF8PROC_DLLEXPORT __declspec(dllexport)
+#    else
+#      define UTF8PROC_DLLEXPORT __declspec(dllimport)
+#    endif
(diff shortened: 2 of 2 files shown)


utf8proc.h
```diff
@@ -113,16 +113,24 @@ typedef bool utf8proc_bool;
...
 #  else
-#    define UTF8PROC_DLLEXPORT __declspec(dllimport)
+#    define UTF8PROC_DLLEXPORT
 #  endif
-#elif __GNUC__ >= 4
-#  define UTF8PROC_DLLEXPORT __attribute__ ((visibility("default")))
 #else
-#  define UTF8PROC_DLLEXPORT
+#  if defined(__GNUC__) && __GNUC__ >= 4
+#    define UTF8PROC_DLLEXPORT __attribute__ ((visibility("hidden")))
```
stevengj: Why would we use `hidden` [visibility](https://gcc.gnu.org/onlinedocs/gcc-6.2.0/gcc/Common-Function-Attributes.html#Common-Function-Attributes) here?

utf8proc.h
```diff
@@ -113,16 +113,24 @@ typedef bool utf8proc_bool;
...
-#    define UTF8PROC_DLLEXPORT __declspec(dllexport)
+#ifndef UTF8PROC_BUILD_AS_LIB
+#  ifdef _WIN32
+#    ifdef UTF8PROC_EXPORTS
+#      define UTF8PROC_DLLEXPORT __declspec(dllexport)
+#    else
+#      define UTF8PROC_DLLEXPORT __declspec(dllimport)
+#    endif
+#  elif defined(__GNUC__) && __GNUC__ >= 4
+#    define UTF8PROC_DLLEXPORT __attribute__ ((visibility("default")))
```
stevengj: If we are building a static library, it doesn't seem like the [visibility attribute](https://gcc.gnu.org/onlinedocs/gcc-6.2.0/gcc/Common-Function-Attributes.html#Common-Function-Attributes) is relevant, since that seems to correspond to the ELF ABI for shared objects.

CMakeLists.txt
```diff
@@ -13,19 +13,55 @@ set(SO_MAJOR 2)
...
+endif(NOT DEFINED _utf8proc_additional_defined)
+
+if(UTF8PROC_BUILD_AS_SHARED)
+  list(APPEND _utf8proc_additional_defined 
+    "-DUTF8PROC_EXPORTS"
+  )
+  set(_utf8proc_build_type "SHARED")
+else(UTF8PROC_BUILD_AS_SHARED)
+  list(APPEND _utf8proc_additional_defined 
+    "-DUTF8PROC_BUILD_AS_LIB"
```
stevengj: `BUILD_AS_STATIC` seems more appropriate as the platform-independent antonym of "shared"

stevengj: Or maybe `UTF8PROC_STATIC_EXPORTS`

CMakeLists.txt
```diff
@@ -13,19 +13,55 @@ set(SO_MAJOR 2)
...
-  -DUTF8PROC_EXPORTS
-)
+# configure build type
+
+set(_utf8proc_is_set_build_shared_libs TRUE)
+if(DEFINED BUILD_SHARED_LIBS)
+  set(_utf8proc_is_set_build_shared_libs ${BUILD_SHARED_LIBS})
+endif(DEFINED BUILD_SHARED_LIBS)
+
+set(UTF8PROC_BUILD_AS_SHARED ${_utf8proc_is_set_build_shared_libs} CACHE BOOL "If present and true, this will cause utf8proc to be built shared unless the utf8proc was explicitly added as a static library.")
```
stevengj: what does "explicitly added as a static library" mean?

CMakeLists.txt
```diff
@@ -13,19 +13,55 @@ set(SO_MAJOR 2)
 set(SO_MINOR 0)
 set(SO_PATCH 2)
 
-add_definitions (
-  -DUTF8PROC_EXPORTS
-)
+# configure build type
+
+set(_utf8proc_is_set_build_shared_libs TRUE)
```
stevengj: This is a weird variable name.

utf8proc.h
nicolaichuk: > All new code should have this support from the beginning!
read more https://gcc.gnu.org/wiki/Visibility

CMakeLists.txt
nicolaichuk: Sorry, my english is not good. Maybe you formulate a better help phrase.

utf8proc.h
Keno: The whole point of these symbols is that they are the public API of this library. Making them hidden makes little sense. End users can decide whether to ask to linker to hide them or not when they build this into their libraries and applications.

CMakeLists.txt
nicolaichuk: You can change to any other :)

utf8proc.h
nicolaichuk: When i build static library - this meen that i want use this library for internal purposes only. And it hide for external user.
When i want external uses - i will build SHARED or MODULE version.

Using visibility("default") for all symbols - is legacy gcc practices.

Keno: > When i build static library - this meen that i want use this library for internal purposes only. And it hide for external user.

It just means that this code is not intended to be used as a shared library (by itself). People would very well decide to use -Wl,--whole-archive and re-export those symbols. Visibility is an API concern and whether or not this library is build shared does not affect what the library's public API is.

nicolaichuk: Maybe.
If you want you can change this to visibility("default").
It is not principal thing for me.

Keno: Yes, just leave that part as-is.

CMakeLists.txt
```diff
@@ -13,19 +13,55 @@ set(SO_MAJOR 2)
...
+if(UTF8PROC_BUILD_AS_SHARED)
+  list(APPEND _utf8proc_additional_defined 
+    "-DUTF8PROC_EXPORTS"
+  )
+  set(_utf8proc_build_type "SHARED")
+else(UTF8PROC_BUILD_AS_SHARED)
+  list(APPEND _utf8proc_additional_defined 
+    "-DUTF8PROC_BUILD_AS_LIB"
+  )
+  set(_utf8proc_build_type "STATIC")
```
nicolaichuk: Also. If you want. You can add support build as MODULE too.

CMakeLists.txt
nicolaichuk: Also. If you want. You can add support build as MODULE too.

utf8proc.h
nicolaichuk: if**N**def UTF8PROC_BUILD_AS_LIB

in this place we are building shared library.

nicolaichuk:
ping...

utf8proc.h
stevengj: I understand why you would want to hide symbols that you are not exporting.  We are already doing this by declaring internal functions as `static` (which works because all of our code is in a single `.c` file). But `UTF8PROC_DLLEXPORT` is used for *exported* functions.  Surely we don't want to hide *those* symbols?

stevengj: Also, this code is for when `UTF8PROC_BUILD_AS_LIB` is defined, so we are not even building a shared library and the visibility attribute is ignored, no?

utf8proc.h
stevengj: My bad; in that case the comment applies to the `visibility("hidden")` above.

stevengj: In any case, this seems a bit superfluous, since the visibility is `default` by default.

stevengj:
You need to update the PR to address the comments above...

erengy:
Just chiming in to say that I'd appreciate an option for static linking. I've been manually editing the header to `#define UTF8PROC_DLLEXPORT` for some time now, so even a simple `ifdef` would help:

```cpp
#ifdef UTF8PROC_STATIC
#  define UTF8PROC_DLLEXPORT
#else
// (current definitions)
#endif
```

DanielSWolf:
Same here. I need static linking, so I'll have to modify your source code. A simple `ifdef` would be great, as suggested by @erengy above and by @madscientist in issue #96.

stevengj:
If anyone feels motivated to fix this PR, that would be great.

maxbrunsfeld:
@stevengj I'm not familiar with cmake, but I'd love to get this working at a basic level. Would you merge a PR that just makes @erengy's suggested change to the header?

stevengj:
Closed by #123.

PR #89, opened 2016-11-29
new utf8proc_map_custom for hooking in user-defined custom mappings

### LINKED ISSUE #11: add custom JULIA normalization?
For JuliaLang/julia#5903.   If utf8proc can have LUMP, then libmojibake can have JULIA.  Unless we want to keep this separate from libmojibake.

stevengj:
This PR supplies two new functions, `utf8proc_map_custom` and `utf8proc_decompose_custom`, which are like the old `utf8proc_map` and `utf8proc_decompose` except that they allow the user to pass a custom mapping function that is applied to every codepoint in the input before any other transformation is performed.

This will allow us to fix #11 without adding any Julia-specific code into utf8proc, and seems like a generally useful mechanism to have.

cc @jiahao, @StefanKarpinski

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -10,8 +10,8 @@ project (utf8proc C)
 # API version number (defined in utf8proc.h).
 # Be sure to also update these in Makefile and MANIFEST!
 set(SO_MAJOR 2)
-set(SO_MINOR 0)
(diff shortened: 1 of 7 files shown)


stevengj:
Good to merge?

PR #92, opened 2016-12-28
Don't use cached version of UnicodeData.txt

MichaelHatherly:
Ref: https://github.com/JuliaLang/julia/pull/19725, UnicodeData.txt is now being cached in JuliaLang/julia's build. cc @tkelman

### DIFF
--- a/data/Makefile
+++ b/data/Makefile
@@ -38,7 +38,7 @@ CharWidths.txt: charwidths.jl unifont.sfd unifont_upper.sfd EastAsianWidth.txt
 UNICODE_VERSION=9.0.0
 
 UnicodeData.txt:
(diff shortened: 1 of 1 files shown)


nalimilan:
Sorry for not commenting earlier, but what's the idea here? Why shouldn't it be cached in utf8proc too, which is an independent project? Also, why is EastAsianWidth.txt still cached?

tkelman:
We're using the same caching server, but the fact that they're separate projects means the unicode versions might not always match. The caching server only saves files by the basename I believe - we could alternately fix it to use the full url in the cached path.

nalimilan:
Got it, thanks. But better do the same for EastAsianWidth.txt, no?

tkelman:
I don't think we download that one for julia, do we?

nalimilan:
No, but if we think it's OK not to cache the first file, why is it necessary to cache the second one? Reading this file without the context, it's quite confusing.

PR #94, opened 2017-01-14
removed inclusion of non-portable header file

H2CO3:
Resolves https://github.com/JuliaLang/utf8proc/issues/93.


stevengj:
Thanks!

PR #102, opened 2017-03-05
Fixes allowing for “Full” folding and NFKC_CaseFold compliance.

nomoon:
Changes:
* Only include `C` (Common) & `F` (Full) foldings from [CaseFolding.txt](http://unicode.org/Public/UCD/latest/ucd/CaseFolding.txt) data. Removed `S` (Simple) since `F` & `S` are specified to be exclusive. I'm not sure if it was causing issues, but it's technically an error.
* Extend `UTF8PROC_IGNORE` to also ignore unassigned codepoints (such as \u2065) which are specified as being discarded by NFKC_CF.

With these changes, compliant NFKC_CaseFolding should be achievable with the options `UTF8PROC_STABLE | UTF8PROC_COMPOSE | UTF8PROC_COMPAT | UTF8PROC_CASEFOLD | UTF8PROC_IGNORE`

_(This PR doesn't include the new `utf8proc_data.c` file since there's no point in generating it multiple times if changes are needed.)_

### DIFF
--- a/data/data_generator.rb
+++ b/data/data_generator.rb
@@ -104,7 +104,7 @@ $excl_version = $excl_version.chomp.split("\n").collect { |e| e.hex }
 $case_folding_string = File.open("CaseFolding.txt", :encoding => 'utf-8').read
 $case_folding = {}
 $case_folding_string.chomp.split("\n").each do |line|
-  next unless line =~ /([0-9A-F]+); [CFS]; ([0-9A-F ]+);/i
+  next unless line =~ /([0-9A-F]+); [CF]; ([0-9A-F ]+);/i
   $case_folding[$1.hex] = $2.split(" ").collect { |e| e.hex }
 end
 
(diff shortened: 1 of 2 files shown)


nomoon:
It appears the tests are failing because I didn't generate and commit the new `utf8proc_data.c`

utf8proc.h
```diff
@@ -690,6 +693,11 @@ UTF8PROC_DLLEXPORT utf8proc_uint8_t *utf8proc_NFC(const utf8proc_uint8_t *str);
 UTF8PROC_DLLEXPORT utf8proc_uint8_t *utf8proc_NFKD(const utf8proc_uint8_t *str);
 /** NFKC normalization (@ref UTF8PROC_COMPOSE and @ref UTF8PROC_COMPAT). */
 UTF8PROC_DLLEXPORT utf8proc_uint8_t *utf8proc_NFKC(const utf8proc_uint8_t *str);
+/**
+ * NFKC_Casefold normalization (@ref UTF8PROC_COMPOSE and @ref UTF8PROC_COMPAT
+ * and @ref UTF8PROC_CASEFOLD and @ref UTF8PROC_IGNORE).
+ **/
+UTF8PROC_DLLEXPORT utf8proc_uint8_t *utf8proc_NFKC_CF(const utf8proc_uint8_t *str);
```
stevengj: Probably just call it `NFKC_Casefold` since that is the standard name.

stevengj:
Closed in favor of the updated PR, #133.

PR #108, opened 2017-07-04
Added cmake code to generate and install a pkgconfig file as well as …

ianwlarson:
…a cmake configure file.


CMakeLists.txt
```diff
@@ -6,6 +6,10 @@ disallow_intree_builds()
 
 project (utf8proc C)
 
+set(INSTALL_LIB_DIR "${CMAKE_INSTALL_PREFIX}/lib" CACHE PATH "Installation director for libraries.")
+set(INSTALL_INC_DIR "${CMAKE_INSTALL_PREFIX}/include" CACHE PATH "Installation directory for headers.")
+set(INSTALL_PKGCONFIG_DIR "${CMAKE_INSTALL_PREFIX}/share/pkgconfig" CACHE PATH "Installation dirctory for pkgconfig file.")
```
tkelman: directory missing the e here, missing the y above

ianwlarson:
The pkgconfig file eases use of the library, and the cmake-config file allows the library to be easily located and accessed from any project using cmake:

```cmake
find_package(utf8proc REQUIRED)
include_directories(${UTF8PROC_INCLUDE_DIRS})
set(CMAKE_C_FLAGS_DEBUG "${CMAKE_C_FLAGS_DEBUG} ${LIBUTF8PROC_DEFINITIONS}")
set(CMAKE_C_FLAGS_RELEASE "${CMAKE_C_FLAGS_RELEASE} ${LIBUTF8PROC_DEFINITIONS}")
```

ianwlarson:
Would you guys consider raising the cmake version requirement to major version 3? Do you guys have any specific reason to keep it at 2.8?

CMakeLists.txt
```diff
@@ -21,13 +25,31 @@ if (NOT MSVC)
   set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -O2 -std=c99 -pedantic -Wall")
 endif ()
 
-add_library (utf8proc
+add_library (utf8proc SHARED
```
stevengj: Wouldn't it be better to use `BUILD_SHARED_LIBS` so that this is configurable?

CMakeLists.txt
```diff
@@ -6,6 +6,10 @@ disallow_intree_builds()
 
 project (utf8proc C)
 
+set(INSTALL_LIB_DIR "${CMAKE_INSTALL_PREFIX}/lib" CACHE PATH "Installation directory for libraries.")
+set(INSTALL_INC_DIR "${CMAKE_INSTALL_PREFIX}/include" CACHE PATH "Installation directory for headers.")
```
stevengj: Are these directories correct for Windows?

CMakeLists.txt
```diff
@@ -21,13 +25,31 @@ if (NOT MSVC)
...
+               ${CMAKE_CURRENT_SOURCE_DIR}/${PROJECT_NAME}-config-version.cmake @ONLY)
+
+install(FILES utf8proc.pc DESTINATION "${INSTALL_PKGCONFIG_DIR}")
+install(FILES utf8proc.h DESTINATION "${INSTALL_INC_DIR}")
+install(TARGETS utf8proc
+        ARCHIVE DESTINATION "${INSTALL_LIB_DIR}"
+        LIBRARY DESTINATION "${INSTALL_LIB_DIR}")
+  
+install(FILES ${PROJECT_NAME}-config.cmake ${PROJECT_NAME}-config-version.cmake
+        DESTINATION "${INSTALL_LIB_DIR}/${PROJECT_NAME}")
```
stevengj: I get the impression that `"${CMAKE_INSTALL_PREFIX}/share/cmake/${PROJECT_NAME}"` is preferred here.  
It makes sense to put it in a `cmake` subdirectory so as not to clutter the `lib` or `share` directory, and it makes sense to use `share` rather than `lib` similar to the `.pc` file.

See e.g. [here](https://stackoverflow.com/questions/30201895/what-is-the-location-to-install-cmake-extension-files-in-ubuntu) and [here](https://cmake.org/cmake/help/v3.2/command/find_package.html).

utf8proc-config-version.cmake.cmakein
```diff
@@ -0,0 +1,11 @@
+set(PACKAGE_VERSION @UTF8PROC_FULL_VERSION@)
```
stevengj: Technically, this should probably be the API version (the one in `utfproc.h` — this is the "human-readable" version number) not the ABI version (the one used for versioning the shared-library file).

(Currently, the two version numbers are the same, but it is possible that they may diverge in the future.)

CMakeLists.txt
tkelman: shared libs should go to bin for windows. cmake usually has the right defaults

CMakeLists.txt
ianwlarson: Good point.

PR #111, opened 2017-09-11
Change: CMake, allow the developer to select if Dynamic or Static lib…

Blonder:
…raries are built


tkelman:
> CMake Version 2.8.6 Now has all of this functionality built in

You don't need to change anything in the cmakelists to do this.

Blonder:
@tkelman Thanks for the info, didn't know that's already build in, I tested the build in feature and it works.

PR #113, opened 2017-09-21
Fix declaration-after-statement warning when compiling in strict C90 mode.

### LINKED ISSUE #112: Compilation fails when C90 mode is enforced
I'm compiling utf8proc in strict C90 mode with `-std=c90 -Werror=declaration-after-statement` (never mind the reasons for that) and get the following error:
```
.../utf8proc.c:292:17: error: ISO C90 forbids mixing declarations and
      code [-Werror,-Wdeclaration-after-statement]
  utf8proc_bool break_permitted = grapheme_break_simple(lbc_override, tbc);
```

brainy:
Fixes #112.


stevengj:
Thanks, LGTM.

Issue #118, opened 2018-02-14
Missing VERSION_INFO when compiling as Windows DLL


Blonder:
The complete version info for the DLL e.g. Company Name, File Version, Program Name, Program Version a.s.o. is missing when compiling as Windows DLL. This leads to an empty Property Page with no information when showing the Windows File Properties Dialog of the DLL (right click onto the DLL and open 'Properties' dialog).

stevengj:
See https://stackoverflow.com/questions/6693100/how-to-generate-windows-dll-versioning-information-with-cmake if someone wants to take this on.

Blonder:
I added another solution with lesser impact and no dependencies (nearly the same as zlib does it) see https://github.com/JuliaStrings/utf8proc/pull/120#issuecomment-423777718

PR #120, opened 2018-03-01
Solved issue #118 "Missing VERSION_INFO when compiling as Windows DLL"

Blonder:
This solves JuliaLang/utf8proc issue #118 "Missing VERSION_INFO when compiling as Windows DLL" using CMakeHelpsers from https://github.com/halex2005/CMakeHelpers. CMakeHelpers must be place in the same directory as / parallel to utf8proc.
For English resource language you need to replace the following things in the CMakeHelpers/VersionResource.rc file:

FILEFLAGSMASK 0x3fL                    with   FILEFLAGSMASK 0x17
LANGUAGE LANG_RUSSIAN, SUBLANG_DEFAULT with   LANGUAGE LANG_ENGLISH, SUBLANG_DEFAULT
BLOCK "041904b0                        with   BLOCK "040704b0"
VALUE "Translation", 0x419, 1200       with   VALUE "Translation", 0x407, 1200

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -2,6 +2,12 @@ cmake_minimum_required (VERSION 2.8)
 
 include (utils.cmake)
 
+set(HASCMAKEHELPERS EXISTS "${CMAKE_SOURCE_DIR}/../CMakeHelpers/generate_product_version.cmake")
+
+if(${HASCMAKEHELPERS})
+  include ("${CMAKE_SOURCE_DIR}/../CMakeHelpers/generate_product_version.cmake")
+endif ()
+
 disallow_intree_builds()
 
 project (utf8proc C)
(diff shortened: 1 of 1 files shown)


stevengj:
I don't like the idea of the build script depending on CMakeHelpers being installed in a particular directory outside the source tree.

CMakeHelpers is MIT-licensed, and https://github.com/halex2005/CMakeHelpers/blob/master/generate_product_version.cmake hasn't changed in 3 years.  Maybe we should just include it in utf8proc?

Blonder:
First I had the same idea to include CMakeHelpers in utf8proc but then I choose the parallel solution because this way I can use it for other projects too where the version info is needed and to be completely independent (no "copy/paste" code that needs to be merged when changes occur - especially important if more than one project uses CMakeHelpers). The CMakeLists.txt file checks if CMakeHelpers is installed and uses it to generate and add the version info or no version info is generated / added.

stevengj:
Bundling a version that is known to work for utf8proc will be much less fragile.

Blonder:
I updated the VERSION_INFO to 2.4.0

CMakeLists.txt
```diff
@@ -10,16 +10,27 @@ project (utf8proc C)
 # API version number (defined in utf8proc.h).
 # Be sure to also update these in Makefile and MANIFEST!
 set(SO_MAJOR 2)
-set(SO_MINOR 3)
-set(SO_PATCH 1)
+set(SO_MINOR 4)
+set(SO_PATCH 0)
```
stevengj: As the comment says, these numbers indicate ABI compatibility and don't follow the API version number exactly.  From 2.3 to 2.4 there was no change in ABI which is why the minor version stayed 3.

stevengj: Though it's probably easier for us to just match the API number so that we don't have this discussion repeatedly.

Blonder:
I refactored the version info problem on Windows and removed the definitions from CMakeLists.txt and moved it to utf8proc.h where the UTF8PROC_VERSION_MAJOR, UTF8PROC_VERSION_MINOR and UTF8PROC_VERSION_PATCH is defined.

Blonder:
Do you have any objects to this new solution?

Issue #96, opened 2017-02-17
Would like to be able to disable DLLEXPORT


madscientist:
In my system I'm linking utf8proc as a static library, into my shared library and I don't want the utf8proc symbols to be visible outside my shared library (in case the user of my shared library has their own version of this library or even other code which has symbol conflicts, in other libraries, etc.)

However currently there's no way to turn off the DLLEXPORT stuff in utf8proc.h.  I'd appreciate it if there were a controlling macro I could add to my CFLAGS that would disable it, maybe something like:

```c
#ifdef UTF8PROC_NO_DLLEXPORT
# define UTF8PROC_DLLEXPORT
#elif defined(_WIN32)
#  ifdef UTF8PROC_EXPORTS
#    define UTF8PROC_DLLEXPORT __declspec(dllexport)
#  else
#    define UTF8PROC_DLLEXPORT __declspec(dllimport)
#  endif
#elif __GNUC__ >= 4
#  define UTF8PROC_DLLEXPORT __attribute__ ((visibility("default")))
#else
#  define UTF8PROC_DLLEXPORT
#endif
```

or something along those lines; I can add `-DUTF8PROC_NO_DLLEXPORT` to CFLAGS to avoid any exporting.

madscientist:
Another alternative is to allow `UTF8PROC_DLLEXPORT` itself to be defined before this and test for that, then I can just set it to the empty string:

```c
#ifndef UTF8PROC_DLLEXPORT
#  ifdef _WIN32
#    ifdef UTF8PROC_EXPORTS
#      define UTF8PROC_DLLEXPORT __declspec(dllexport)
#    else
#      define UTF8PROC_DLLEXPORT __declspec(dllimport)
#    endif
#  elif __GNUC__ >= 4
#    define UTF8PROC_DLLEXPORT __attribute__ ((visibility("default")))
#  else
#    define UTF8PROC_DLLEXPORT
#  endif
#endif
```

vtjnash:
see PR https://github.com/JuliaLang/utf8proc/pull/85

madscientist:
Thanks for that reference @vtjnash 

[...]

PR #122, opened 2018-04-06
`#define UTF8PROC_STATIC` to disable DLLEXPORT

past-due:
This is a simplified subset of #85 (also, see: #96) that enables compiling utf8proc without DLLEXPORT - without requiring the library consumer to modify the source files.

Simply `#define UTF8PROC_STATIC` when compiling to disable DLLEXPORT.


stevengj:
Closed by #123.

PR #136, opened 2018-05-04
allow installation in nonstandard directories

GitMensch:
copied from https://raw.githubusercontent.com/lz4/lz4/dev/lib/Makefile, tested - works

### DIFF
--- a/README.md
+++ b/README.md
@@ -30,9 +30,21 @@ data governed by the similarly permissive [Unicode data
 license](http://www.unicode.org/copyright.html#Exhibit1)); please see
 the included `LICENSE.md` file for more detailed information.
 
-## Quick Start
+## Installation
 
-For compilation of the C library run `make`.
+```
+make            # compilation of the C library
+make install    # this command may require root access
+```
+
+utf8proc's `Makefile` supports standard [Makefile conventions],
+including [staged installs], [redirection], or [command redefinition].
+It is compatible with parallel builds (`-j#`).
+
+[Makefile conventions]: https://www.gnu.org/prep/standards/html_node/Makefile-Conventions.html
+[staged installs]: https://www.gnu.org/prep/standards/html_node/DESTDIR.html
+[redirection]: https://www.gnu.org/prep/standards/html_node/Directory-Variables.html
+[command redefinition]: https://www.gnu.org/prep/standards/html_node/Utilities-in-Makefiles.html
 
 ## General Information
 


Makefile
```diff
@@ -33,9 +33,17 @@ else # GNU/Linux, at least (Windows should probably use cmake)
 endif
 
 # installation directories (for 'make install')
-prefix=/usr/local
-libdir=$(prefix)/lib
-includedir=$(prefix)/include
+DESTDIR     ?=
+# directory variables : GNU conventions prefer lowercase
+# see https://www.gnu.org/prep/standards/html_node/Makefile-Conventions.html
+# support both lower and uppercase (BSD), use uppercase in script
```
stevengj: You introduced the uppercase variables, but didn't actually use them in the `install` target below

stevengj:
Doesn't e.g. `make prefix=/foo/bar` already work prior to this patch?  Setting a variable on the command line should already [override that variable in the file](https://www.gnu.org/software/make/manual/html_node/Overriding.html) even without using `?=`.

`?=` only allows the variable to be overridden by an environment variable, but you can also do that by just running `make -e`.

GitMensch:
Doesn't e.g. `make prefix=/foo/bar` already work prior to this patch?

No, `make prefix=/foo/bar install` was actually what I've did - and got an access denied for `/usr/something`. After using the Makefile including in the PR the install worked.

Makefile
nalimilan: Better have only the lowercase variants, things get really messy with two variables for each directory.

nalimilan:
I'm surprised to hear it doesn't work, as I use `make install DESTDIR=%{buildroot} prefix=%{_prefix} includedir=%{_includedir} libdir=%{_libdir}` in the Fedora package and it works fine.

stevengj:
> and got an access denied for `/usr/something`

An "access denied" error has nothing to do with the Makefile.  You just forgot to use `sudo`.

GitMensch:
> An "access denied" error has nothing to do with the Makefile. You just forgot to use `sudo`.

No, this has something to do with the Makefile as it was raised when doing

`make prefix=/foo/bar install` - I did not intended to install anything into system space.

With the changed Makefile the same command worked perfectly fine...

stevengj:
@GitMensch, what version of `make` are you using?  (What is `make -v`)

GitMensch:
@stevengj 

GNU Make 4.2.1
Built for i686-pc-linux-gnu

To finish this PR:
Do you prefer to drop the "BSD variant" (upper-case variables) completely or change the Makefile to use the new defined upper-case variables (that will be set from the upper-case variables when passed to make, otherwise from the lower-case variables when passed to make, otherwise from the old default)?
I'll commit the appropriate change.

stevengj:
I'm still trying to figure out what problem is being solved here.  The current `Makefile` already has a `prefix` variable.   `make prefix=foo` should already override this in GNU make.   The main difference with `?=`, as I understand it, arises if e.g. you include our Makefile in another makefile and you want the former's definition to take precedence, but that isn't the case here.

PR #138, opened 2018-06-02
Enhance CMakeLists.txt

nehaljwani:
- Add install target so that one can `nmake install` it in a specified `CMAKE_INSTALL_PREFIX`
- Suffix static library with '_static' in case the compiler is `MSVC`. Useful for packaging both shared and static libraries in Windows in a single tarball.


stevengj:
Why does Windows need a `_static` suffix, since it already has `.lib` vs. `.dll`?  Is that because there is also a static import library for the dll?

nehaljwani:
Yeah, for the shared dll, there is a .lib import file which gets created, which conflicts with the name of the static .lib

Wohlstand:
Usually when you building both shared and static builds of the library

Issue #105, opened 2017-05-14
Add a pkgconfig file


hobbitalastair:
The netsurf project [uses](http://source.netsurf-browser.org/netsurf.git/tree/Makefile#n532) a [pkgconfig file](http://source.netsurf-browser.org/libutf8proc.git/tree/libutf8proc.pc.in) to find linking flags for utf8proc. It would be nice to have this in the upstream library.

@tlsa this seems to be required to build against upstream instead of netsurf's fork.

stevengj:
Seems reasonable to add.

ianwlarson:
[utf8proc.txt](https://github.com/JuliaLang/utf8proc/files/1123089/utf8proc.txt)

Very straightforward to make.

ianwlarson:
#108

stevengj:
CLosed by #142.

PR #142, opened 2018-08-23
Generate and install a pkg-config file.

tlsa:
This adds a pkg-config file for libutf8proc.

Fixes issue https://github.com/JuliaStrings/utf8proc/issues/105.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -34,8 +34,8 @@ endif
 
 # installation directories (for 'make install')
 prefix=/usr/local
-libdir=lib
-includedir=include
+libdir=$(prefix)/lib
+includedir=$(prefix)/include
 pkgconfigdir=$(libdir)/pkgconfig
 
 # meta targets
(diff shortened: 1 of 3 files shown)


tlsa:
@stevengj Does this look reasonable?

Makefile
```diff
@@ -34,8 +34,9 @@ endif
 
 # installation directories (for 'make install')
 prefix=/usr/local
-libdir=$(prefix)/lib
-includedir=$(prefix)/include
+libdir=lib
```
stevengj: In the [GNU-style installation variables](https://www.gnu.org/prep/standards/html_node/Directory-Variables.html), `libdir` should be the whole path, defaulting to `$(prefix)/lib`.   So the previous version of this code was correct.

tlsa:
@stevengj Oh, good point.  I still think it's worth ensuring the pkgconfig files stay in sync though, so I've reverted the previous change, and done it like this https://github.com/JuliaStrings/utf8proc/pull/142/commits/53fb330c59c7f1baaa49a10b2228f96bd7b80e6f instead.

Makefile
```diff
@@ -36,6 +36,10 @@ endif
 prefix=/usr/local
 libdir=$(prefix)/lib
 includedir=$(prefix)/include
+pkgconfigdir=$(libdir)/pkgconfig
+
+pkglibdir=$(libdir:$(prefix)/%=%)
```
nalimilan: I don't understand why you compute these variables. Why not use `libdir` directly in the pkgconfig file, and hardcode the prefix in each variable? I don't see the advantage of using `${prefix}` there.

tlsa: Here are some examples:

* https://people.freedesktop.org/~dbn/pkg-config-guide.html
* https://linux.die.net/man/1/pkg-config
* https://en.wikipedia.org/wiki/Pkg-config

nalimilan: I see that, but what's the point when it makes the code more complex?

tlsa: When you run pkg-config, e.g. to get the appropriate cflags for compiling the library, you can use stuff like `pkg-config --define-prefix --cflags libpng`, which will change what is substituted for `$(prefix)` when it spits out the includedir.

nalimilan: OK, makes sense then!

tlsa:
@stevengj Is this OK now?

PR #152, opened 2019-04-17
[CMake] Add UTF8PROC_NO_INSTALL option

past-due:
In some cases (such as using the library via `add_subdirectory` in CMake), it may be desirable to disable installation.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -13,7 +13,7 @@ set(SO_MAJOR 2)
 set(SO_MINOR 2)
 set(SO_PATCH 0)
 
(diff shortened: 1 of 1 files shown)


stevengj:
Thanks, seems reasonable, though it would be better if cmake provided a cleaner way to do this from the parent (as is discussed in various places online, e.g. https://stackoverflow.com/questions/35344132/cpack-exclude-install-commands-from-subdirectory-googletest-directory/35345933, glfw/glfw#30, ...)

I modifed the PR so that it is `UTF8PROC_INSTALL=No`, however.

PR #158, opened 2019-05-20
Add meson compilation system

yageek:
As I required to use this library inside a meson based project,
I translated the `CMake` version of this project.

### DIFF
--- a/README.md
+++ b/README.md
@@ -43,6 +43,14 @@ cmake ..
 make
 ```
 
+or by using `meson`:
+```sh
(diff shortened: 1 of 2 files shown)


stevengj:
There are too many build systems out there for us to add them all to utf8proc.

To use utf8proc in a project built with a different build system, you have two options:

1. Treat it like any other dependency, and require libutf8proc to be installed in an appropriate location.

2. Have your build system call `make` or `cmake` to build libutf8proc in a subdirectory.

PR #161, opened 2019-07-25
Fix 2 excessive dependencies in Makefile

vemakereporter:
Hi, I've fixed 2 excessive dependencies reported.
Those issues can cause unnecessary rebuilds when `utf8proc` is incrementally built.

For example, any changes in `data/charwidths.jl` will the unnecessary rebuild of `data/utf8proc_data.c.new`, which can slow down the build process of the total project.

I modify the Makefile to remove these excessive dependencies. I've checked and the new Makefile works well.
Looking forward to your confirmation.

Thanks
Vemake

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -70,7 +70,7 @@ manifest: MANIFEST.new
 
 # real targets
 
-data/utf8proc_data.c.new: libutf8proc.$(SHLIB_EXT) data/data_generator.rb data/charwidths.jl
+data/utf8proc_data.c.new: libutf8proc.$(SHLIB_EXT) data/data_generator.rb
 	$(MAKE) -C data utf8proc_data.c.new
 
 utf8proc.o: utf8proc.h utf8proc.c utf8proc_data.c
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -70,7 +70,7 @@ manifest: MANIFEST.new
 
 # real targets
 
-data/utf8proc_data.c.new: libutf8proc.$(SHLIB_EXT) data/data_generator.rb data/charwidths.jl
+data/utf8proc_data.c.new: libutf8proc.$(SHLIB_EXT) data/data_generator.rb
```
stevengj: This dependency is correct. The data file depends on charwidths.jl

Makefile
```diff
@@ -88,7 +88,7 @@ libutf8proc.so: libutf8proc.so.$(MAJOR).$(MINOR).$(PATCH)
 	ln -f -s libutf8proc.so.$(MAJOR).$(MINOR).$(PATCH) $@
 	ln -f -s libutf8proc.so.$(MAJOR).$(MINOR).$(PATCH) $@.$(MAJOR)
 
-libutf8proc.$(MAJOR).dylib: utf8proc.o
+libutf8proc.$(MAJOR).dylib:
```
stevengj: This dependency is correct: the library is built from utf8proc.o. `$^` refers to this.

stevengj:
Both of these dependencies are correct as-is.

Issue #170, opened 2020-01-27
utf8proc_charwidth returns 1 for é character


markand:
Hi,

I'm sorry if I misunderstood the purpose of utf8proc_charwidth, I thought it'd return the number of bytes required for a given codepoint.

Example, the letter é in UTF-8 is two bytes long (value of 233 in unicode) but `utf8proc_charwidth(233)` always return 1.

Did I miss something or there is no function to get the number of bytes a codepoint require?

PallHaraldsson:
I think you have this in mind:

```
julia> ncodeunits("é")
2
```

I think the other has to do with with on the screen, if I recall to support Chinese.

stevengj:
charwidth is an approximate on-screen width in fixed-width fonts, not the number of code units.

PR #172, opened 2020-01-28
add new utf8proc_codepoint_units function

markand:
I propose a simple function to determine the number of UTF-8 bytes that a codepoint require. This mostly addresses the issue #170 

Example:

```c
utf8proc_codepoint_units(233)    // é -> 2
utf8proc_codepoint_units(0x20ac) // € -> 3
```


stevengj:
Maybe `utf8proc_sizeof_char`?

markand:
Hmm, the term sizeof in C has special meaning, here the functions is meant to obtain a variable number between 0 and 4 which to me is the opposite of what "sizeof" sounds to me which always evaluates to a constant object size.

Other alternatives:

- utf8proc_codepoint_utf8len
- utf8proc_codepoint_mblength (multibyte length)
- utf8proc_codepoint_length (pretty name but still confusing at a glance IMHO)

stevengj:
sizeof refers to a size in bytes, which is why it seems appropriate here.

markand:
Updated.

test/sizeofchar.c
```diff
@@ -0,0 +1,41 @@
...
+
+int main(int argc, char **argv)
+{
+     int c, error = 0;
+
+     (void) argc; /* unused */
+     (void) argv; /* unused */
+
+     /* some simple sanity tests of  */
+     for (c = 0; c < 0x80; c++) {
```
stevengj: It would be good to check consistency with `utf8proc_encode_char`, i.e. check `utf8proc_sizeof_char(c) == utf8proc_encode_char(c, buf)` where you declare `utf8proc_uint8_t buf[4]`.

stevengj:
Not sure what the Travis failure is here?

PR #173, opened 2020-01-29
Add: tests to CMakeLists.txt

Andreas-Schniertshauer:


stevengj:
Can you update Travis and AppVeyor so that they exercise the `cmake` tests?

stevengj:
We should probably use `fgets` rather than `getline`.

Andreas-Schniertshauer:
> 
> 
> Can you update Travis and AppVeyor so that they exercise the `cmake` tests?

Done.

CMakeLists.txt
```diff
@@ -64,30 +65,32 @@ if(UTF8PROC_ENABLE_TESTING)
...
   add_test(testcase case)
-#  add_test(testcharwidth charwidth)
   add_test(testcustom custom)
-#  add_test(testgraphemetest graphemetest GraphemeBreakTest.txt)
   add_test(testiterate iterate)
   add_test(testmisc misc)
-#  add_test(testnormtest normtest)
   add_test(testprintproperty printproperty)
   add_test(testvalid valid)
+  if (NOT MSVC)
```
stevengj: shouldn't this use CheckSymbolExists rather than looking at the compiler name?

CMakeLists.txt
```diff
@@ -59,3 +60,37 @@ if (UTF8PROC_INSTALL)
...
+    # Some test disabled, because they don't compile on Windows, missing getline, ...
+    add_executable(charwidth test/tests.h test/tests.c utf8proc.h test/charwidth.c)
+    target_link_libraries(charwidth utf8proc)
+    #add_executable(graphemetest test/tests.h test/tests.c utf8proc.h test/graphemetest.c)
+    #target_link_libraries(graphemetest utf8proc)
+    #add_executable(normtest test/tests.h test/tests.c utf8proc.h test/normtest.c)
+    #target_link_libraries(normtest utf8proc)
+    add_test(testcharwidth charwidth)
+    # add_test(testgraphemetest graphemetest GraphemeBreakTest.txt)
+    # add_test(testnormtest normtest)
```
stevengj: Why are these commented out?   You call them as
```
	test/normtest data/NormalizationTest.txt
	test/graphemetest data/GraphemeBreakTest.txt
```
but then you need a makefile rule to download these two data files.

CMakeLists.txt
Andreas-Schniertshauer: I used if (NOT WIN32) to temporary disable compilation / executing of some test, because the did not compile on Windows because of missing getline, ... This should later be removed when all tests compile.

CMakeLists.txt
Andreas-Schniertshauer: It's commented out because the .txt files are missing, because in CMakeLists.txt is missing building these.

stevengj:
Thanks!  Additional improvements (e.g. removing `getline` or downloading the `.txt` file in cmake) can be done in a subsequent PR.

PR #174, opened 2020-02-29
Build shared library for Windows in Makefile

giordano:
I hope I didn't miss anything

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -29,17 +29,20 @@ PATCH=1
 OS := $(shell uname)
 ifeq ($(OS),Darwin) # MacOS X
   SHLIB_EXT = dylib
-  SHLIB_VERS_EXT = $(MAJOR).dylib
+  SHLIB_VERS_EXT = .$(MAJOR).$(SHLIB_EXT)
+else ifneq (,$(findstring MSYS_NT,$(OS))) # MinGW
+  SHLIB_EXT = dll
+  SHLIB_VERS_EXT = -$(MAJOR).$(SHLIB_EXT)
 else # GNU/Linux, at least (Windows should probably use cmake)
   SHLIB_EXT = so
-  SHLIB_VERS_EXT = so.$(MAJOR).$(MINOR).$(PATCH)
+  SHLIB_VERS_EXT = .$(SHLIB_EXT).$(MAJOR).$(MINOR).$(PATCH)
 endif
 
 # installation directories (for 'make install')
 prefix=/usr/local
 libdir=$(prefix)/lib
 includedir=$(prefix)/include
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -29,17 +29,20 @@ PATCH=1
 OS := $(shell uname)
 ifeq ($(OS),Darwin) # MacOS X
   SHLIB_EXT = dylib
-  SHLIB_VERS_EXT = $(MAJOR).dylib
+  SHLIB_VERS_EXT = .$(MAJOR).$(SHLIB_EXT)
+else ifneq (,$(findstring MSYS_NT,$(OS))) # MinGW
```
Keno: I don't think this is sufficient. Probably needs to copy the OS normalization from Julia

giordano: This: https://github.com/JuliaLang/julia/blob/2f8cdd045cc6b7e77db14393e39cfee8ef4d4a66/Make.inc#L357-L368?

Makefile
```diff
@@ -102,17 +121,19 @@ libutf8proc.pc: libutf8proc.pc.in
...
-	$(INSTALL) -m 755 libutf8proc.$(SHLIB_VERS_EXT) $(DESTDIR)$(libdir)
+	$(INSTALL) -m 755 libutf8proc$(SHLIB_VERS_EXT) $(DESTDIR)$(libdir)
 	mkdir -m 755 -p $(DESTDIR)$(pkgconfigdir)
 	$(INSTALL) -m 644 libutf8proc.pc $(DESTDIR)$(pkgconfigdir)/libutf8proc.pc
-	ln -f -s libutf8proc.$(SHLIB_VERS_EXT) $(DESTDIR)$(libdir)/libutf8proc.$(SHLIB_EXT)
-ifneq ($(OS),Darwin)
-	ln -f -s libutf8proc.$(SHLIB_VERS_EXT) $(DESTDIR)$(libdir)/libutf8proc.so.$(MAJOR)
+	ln -f -s libutf8proc$(SHLIB_VERS_EXT) $(DESTDIR)$(libdir)/libutf8proc.$(SHLIB_EXT)
+ifneq ($(OS), WINNT)
+  ifneq ($(OS),Darwin)
```
Keno: I think in Make conditionals only work at top level?

giordano: They should work also inside a rule, if this is what you mean: https://www.gnu.org/software/make/manual/html_node/Conditional-Example.html.  Or are you asking about the nested conditionals?

stevengj:
Can you add a test to appveyor?

stevengj:
I think Windows users should probably just use `cmake` these days…

PR #177, opened 2020-03-27
Tests with prefix utf8proc.

Andreas-Schniertshauer:
I added the prefix utf8proc. to all tests to be able to see the library name when using grep Passed or Failed and multiple libraries.


.gitignore
```diff
@@ -28,3 +28,7 @@ test/iterate
 test/case
 test/custom
 /tmp/
+mingw_static
+mingw_shared
+msvc_shared
+msvc_static
```
stevengj: Probably these should not be here?

Andreas-Schniertshauer: When compiling with AppVeyor see appveyor.yml, these directories are created, but you don't want to see it in git, because you you will not commit the files in these folders.

stevengj: I'm not sure we care about what git sees on AppVeyor, but fine.

stevengj: (Similarly, we don't add `build_shared` even though we use this directory on Travis.)

PR #178, opened 2020-03-27
Download data and execute commented out tests

Andreas-Schniertshauer:
Download missing data and enabled executing tests that depend on it.


.gitignore
```diff
@@ -28,3 +28,7 @@ test/iterate
 test/case
 test/custom
 /tmp/
+mingw_static
+mingw_shared
+msvc_shared
+msvc_static
```
stevengj: rm

CMakeLists.txt
```diff
@@ -1,96 +1,118 @@
...
-    add_executable(graphemetest test/tests.h test/tests.c utf8proc.h test/graphemetest.c)
-    target_link_libraries(graphemetest utf8proc)
-    add_executable(normtest test/tests.h test/tests.c utf8proc.h test/normtest.c)
-    target_link_libraries(normtest utf8proc)
-    add_test(testcharwidth charwidth)
-    #add_test(testgraphemetest graphemetest data/GraphemeBreakTest.txt)
-    #add_test(testnormtest normtest data/NormalizationTest.txt)
-  endif()
-endif()
+cmake_minimum_required (VERSION 2.8.12)
```
stevengj: Why is this huge diff showing even though most of these lines weren't changed?

Andreas-Schniertshauer: sorry, seems to be a line ending problem Unix <-> Windows.

stevengj:
I'd rather use the built-in `file(DOWNLOAD url ...)` as I commented in #180.   Progress output is pretty unimportant here because the files are so small, and I've had too many bad experiences with broken `curl` installations on user machines.

CMakeLists.txt
```diff
@@ -63,6 +63,27 @@ endif()
...
+      COMMAND ${curl_cmd} ${curl_param} -o data/EastAsianWidth.txt -O http://www.unicode.org/Public/${UNICODE_VERSION}/ucd/EastAsianWidth.txt
+      COMMAND ${curl_cmd} ${curl_param} -o data/GraphemeBreakProperty.txt -O http://www.unicode.org/Public/${UNICODE_VERSION}/ucd/auxiliary/GraphemeBreakProperty.txt
+      COMMAND ${curl_cmd} ${curl_param} -o data/DerivedCoreProperties.txt -O http://www.unicode.org/Public/${UNICODE_VERSION}/ucd/DerivedCoreProperties.txt
+      COMMAND ${curl_cmd} ${curl_param} -o data/CompositionExclusions.txt -O http://www.unicode.org/Public/${UNICODE_VERSION}/ucd/CompositionExclusions.txt
+      COMMAND ${curl_cmd} ${curl_param} -o data/CaseFolding.txt -O http://www.unicode.org/Public/${UNICODE_VERSION}/ucd/CaseFolding.txt
+      COMMAND ${curl_cmd} ${curl_param} -o data/NormalizationTest.txt -O http://www.unicode.org/Public/${UNICODE_VERSION}/ucd/NormalizationTest.txt
+      COMMAND ${curl_cmd} ${curl_param} -o data/GraphemeBreakTestOrg.txt -O http://www.unicode.org/Public/${UNICODE_VERSION}/ucd/auxiliary/GraphemeBreakTest.txt
```
stevengj: You don't need to download all of these.  Only `NormalizationTest.txt` and `GraphemeBreakTest.txt` are needed for the tests.

(The other files are only needed for re-generating the Unicode data tables, which will be unnecessary in `cmake` for the foreseeable future.)

CMakeLists.txt
```diff
@@ -63,6 +63,12 @@ endif()
 
 if(UTF8PROC_ENABLE_TESTING)
   enable_testing()
+  file(MAKE_DIRECTORY data)
+  set(UNICODE_VERSION 13.0.0)
+  file(DOWNLOAD http://www.unicode.org/Public/${UNICODE_VERSION}/ucd/NormalizationTest.txt data/NormalizationTest.txt SHOW_PROGRESS)
+  file(DOWNLOAD http://www.unicode.org/Public/${UNICODE_VERSION}/ucd/auxiliary/GraphemeBreakTest.txt data/GraphemeBreakTestOrg.txt SHOW_PROGRESS)
+  file(REMOVE data/GraphemeBreakTestOrg.txt)
```
stevengj: Why not download as `data/GraphemeBreakTest.txt`?

Andreas-Schniertshauer: Is the same name as in the Makefile and data\Makefile

Andreas-Schniertshauer: Woodhead error, removed a line `    execute_process(COMMAND bash -c "cat data/GraphemeBreakTestOrg.txt | /usr/bin/perl -pe 's,÷,/,g;s,×,+,g'" OUTPUT_FILE data/GraphemeBreakTest.txt)
`

stevengj:
Weird, I thought the tests were green when I merged, and now the graphemes test is timing out on master?

stevengj:
Looks like the download doesn't happen under the msys makefile build, so the tests hang.  For now. I'll just remove the `ctest` for MinGW on AppVeyor.

Andreas-Schniertshauer:
I think it's the execute_process line, where perl is used to strip lines from the file and without stripping these lines grapehme runs into an endless loop.

Andreas-Schniertshauer:
The perl command replaces ÷ with and × with + in the file `data/GraphemeBreakTestOrg.txt`, so maybe we can use the cmake `file(STRINGS...` command and a loop over the lines to replace the chars.

Andreas-Schniertshauer:
I came up with this function instead of the perl / execute_process call:
`
  function(replaceinplace IN_FILE)
    # create list of lines form the contens of a file
    file(STRINGS ${IN_FILE} LINES NEWLINE_CONSUME ENCODING UTF-8)
  
    # overwrite the file....
    file(WRITE ${IN_FILE} "")
  
    # loop through the lines,
    # remove unwanted parts 
    # and write the (changed) line ...
    foreach(LINE IN LISTS LINES)
      string(REPLACE "÷" "/" REPLACED1 "${LINE}")
      string(REPLACE "×" "+" REPLACED2 "${REPLACED1}")
      file(APPEND ${IN_FILE} "${REPLACED2}")
    endforeach()
  endfunction()
`
it works fine on Linux but not on Windows.

PR #180, opened 2020-03-27
Windows10 curl

Andreas-Schniertshauer:
Add: Windows 10 downloading of data using C:\Windows\system32\curl.exe, now the only thing we need to run all tests on Windows is to replace wcwidth and getline in the code to enable the commented out tests.


stevengj:
Can't one use [`file(DOWNLOAD url ...)`](https://cmake.org/cmake/help/v3.12/command/file.html)?

Andreas-Schniertshauer:
Yes you can use it, but the progress output is not as nice as with curl.
If there is no progress output one might think that there is a problem in the script because it takes some seconds to download all files and without progress there is no feedback to the user in this time.

stevengj:
Please consolidate any changes into #178, and use the built-in `file(DOWNLOAD url ...)` rather than adding a dependency on `curl`.

PR #186, opened 2020-03-29
Mingw problems with grapheme test

Andreas-Schniertshauer:
using function replaceinplace instead of execute_process and perl to run tests on MIN_GW (don't know if this works, could not test it locally)


stevengj:
What exactly is the problem that requires the file to be modified?

stevengj:
Probably a cleaner alternative would be to modify the grapheme test so that it can handle the `÷` and `×` characters directly, so that the `perl` transformation can be removed.

Andreas-Schniertshauer:
> 
> 
> What exactly is the problem that requires the file to be modified?
I did not know - haven't analyzed it, I used the same code in the CMakeLists.txt as the code I found in the original Makefile.

stevengj:
In 11bb3d9dc796bb006c79c2962a7d19abcadfb3df I fixed the test so that it works on the unmodified data file, so hopefully we can just re-enable the test if that was the source of the problem.

PR #189, opened 2020-03-29
Fix memory leaks in tests case.c and misc.c

Andreas-Schniertshauer:
Fix: memory leaks in tests case.c and misc.c forgot to call free after calling utf8proc_NFKC_Casefold


stevengj:
Sure, this is fine.

(Memory "leaks" in short-running tests like this aren't actually a practical problem, because the OS frees the memory when the program exits.  But I guess if you are using debugging tools to look for leaks in libutf8proc itself they might be annoying.)

PR #196, opened 2020-07-10
add islower/isupper functions

### LINKED ISSUE #195: Include case information from DerivedCoreProperties.txt
Hi,

Roman letter Ⅰ to Ⅿ and circled letters Ⓐ to Ⓩ are upper case characters, which should be listed in DerivedCoreProperties.txt, and many more for lower case codepoints. Are there plans to include this information?

Regards,

Maarten

ref: https://github.com/apache/arrow/pull/7656

stevengj:
Closes #195.

@maartenbreddels, this adds `islower` and `isupper` functions that should correspond to the `Lowercase` and `Uppercase` derived properties.   I figured out a trick to do it without expanding the data tables, just slightly tweaking the table entries of lower/uppercase characters that don't yet have upper/lowercase mappings, respectively.

~~I still need to add a test.~~


maartenbreddels:
Very cool, if you give me some time, I could test this with https://github.com/apache/arrow/pull/7656

stevengj:
Added some tests and fixed a bug.  This seems to be working (it returns true for `isupper`/`islower` if and only if the character is listed as such in `DerivedProperties.txt`).

PR #197, opened 2020-07-29
Added HPP generator

J-Vernay:
**What?**
This pull request contains a generator (python3 script) to parse utf8prop.c and convert it to a constexpr C++ header-only library.
The resulting library is placed in the `utf8proc` namespace, and `utf8proc_` prefixes are removed.
Private entities are in the namespace `utf8proc::detail`.

**Why?**
In C++, constexpr allows programmers to evaluate functions at compile-time (before the linker).
This is a big part of C++ meta-programming.
I could not find a constexpr library for Unicode, so here is my solution.
I used `utf8proc` as a basis because I do not want to implement these algorithms myself, and other libraries such as `ICU` are way too big for me to dive into and produce a generator in a fairly short period of time.

**Modifications on existing code?**
- Addition of a preprocessor command: UTF8PROC_NO_ALLOC.
  When defined, it removes the functions that are calling _malloc/realloc_.
  It was needed because such functions cannot be marked `constexpr`.
- Enforcement of initialization of all variables (needed for constexpr).
- Modification of an internal type: `static const char s[][3]` -> `static const char * const s[]`
  This does not modify the syntax anywhere, but it changes the string storage from
  static storage to literal storage (needed in constexpr because static storage is not allowed).
All the tests of the C library passes correctly after these modifications.

**License**
The generation script is under the Boost Software License 1.0.
This is a more permissive license than the "expat" license:
The only requirement is to let the copyright notice inside the `gen_hpp.py` script.
Nowhere else.


LB--:
I've been wanting a way to build utf8proc as a constexpr lib as well, but this PR is utterly massive and I don't like this codegen approach to it. I think it should be possible to just define a valueless macro and include the .c file in a C++ source file to enable constexpr support. No need for C++ namespaces either.

stevengj:
What is the motivation for making a constexpr version of utf8proc?  Do you have benchmarks of a real application where it is beneficial?

I would think that most realistic applications of utf8proc would apply its functions mainly to strings only available at runtime, and hence not benefiting much from compile-time evaluation.

LB--:
It's not about performance, it's about being able to normalize embedded strings during compile time. Alternatives are preprocessing (e.g. host build tool step, git hooks/filters) or normalizing during runtime (e.g. during global variable initialization or first use). Both have undesirable drawbacks (build system/source control complications, unnecessary runtime performance loss, etc). For strings known at build time anyway it's simpler to just make the compiler double as the preprocessor for them, using the same code as what runtime would use.

stevengj:
> It's not about performance, it's about being able to normalize embedded strings during compile time. 

Immediately afterwards, however, you identify the main drawback of normalizing at runtime as "unnecessary runtime performance loss".   In other words, it *is* about performance?

Can you sketch a realistic application where static normalization is important?   (I would think that most real applications working with normalization-sensitive Unicode strings would want to load them at runtime anyway for [i18n/l10n](https://en.wikipedia.org/wiki/Internationalization_and_localization) purposes.)

PR #201, opened 2020-10-09
docs: fix simple typo, encounted -> encountered

timgates42:
There is a small typo in utf8proc.h.

Should read `encountered` rather than `encounted`.


stevengj:
Thanks!

PR #202, opened 2020-10-09
slightly improve utf8proc_encode_char()

jakwings:
* add more comments

* use bitwise OR calculation instead of addition


stevengj:
Is bitwise-or faster than + here, or is this just for clarity?

iology:
It's mostly for clarity.  I think using bitwise-or is more intuitive for bit operation here and the addition `+` is outstanding here for no good reason.  I believe without carry `|` is generally faster than `+` though both of them are single instructions.  I was just randomly looking at the source code and not benchmarking on this common bitwise operation. :)

stevengj:
It would be good to benchmark this just to be sure, since this is an inner-loop function.

PR #204, opened 2020-11-19
Fix grapheme breaks on string-initial repeated regional flags and emoji with extended+zwj sequence

### LINKED ISSUE #203: Grapheme breaks
Grapheme breaks seem to be wrong. The sequence ಇಲ್ಲಿ should be two graphemes, yet utf8proc sees it as three 
image

archermarx:
Fixes #203.   https://github.com/JuliaLang/julia/issues/37680 will also be fixed once Julia updates utf8proc. 

@kdheepak wrote the patches, I added some tests

### DIFF
--- a/utf8proc.h
+++ b/utf8proc.h
@@ -502,7 +502,7 @@ UTF8PROC_DLLEXPORT utf8proc_ssize_t utf8proc_decompose_char(
  * string and orders the decomposed sequences correctly.
  *
  * If the @ref UTF8PROC_NULLTERM flag in `options` is set, processing
- * will be stopped, when a NULL byte is encounted, otherwise `strlen`
+ * will be stopped, when a NULL byte is encountered, otherwise `strlen`
  * bytes are processed.  The result (in the form of 32-bit unicode
  * codepoints) is written into the buffer being pointed to by
  * `buffer` (which must contain at least `bufsize` entries).  In case of


utf8proc.c
```diff
@@ -301,6 +301,8 @@ static utf8proc_bool grapheme_break_extended(int lbc, int tbc, utf8proc_int32_t
     // forbidden by a different rule such as GB9).
     if (*state == tbc && tbc == UTF8PROC_BOUNDCLASS_REGIONAL_INDICATOR)
       *state = UTF8PROC_BOUNDCLASS_OTHER;
+    else if (*state == UTF8PROC_BOUNDCLASS_START && lbc == UTF8PROC_BOUNDCLASS_REGIONAL_INDICATOR && tbc == UTF8PROC_BOUNDCLASS_REGIONAL_INDICATOR)
```
stevengj, line 304: Probably this should be merged with the previous condition?
```c
if (tbc == UTF8PROC_BOUNDCLASS_REGIONAL_INDICATOR && 
    (*state == tbc || (*state == UTF8PROC_BOUNDCLASS_START && lbc == UTF8PROC_BOUNDCLASS_REGIONAL_INDICATOR)))
    *state = UTF8PROC_BOUNDCLASS_OTHER;
```

archermarx: Done, tests pass on my machine

utf8proc.c
```diff
@@ -310,6 +311,8 @@ static utf8proc_bool grapheme_break_extended(int lbc, int tbc, utf8proc_int32_t
       else
         *state = tbc;
     }
+    else if (*state == UTF8PROC_BOUNDCLASS_EXTEND && tbc == UTF8PROC_BOUNDCLASS_ZWJ)
```
stevengj, line 314: Why are we in `UTF8PROC_BOUNDCLASS_EXTEND` at this point rather than `UTF8PROC_BOUNDCLASS_EXTENDED_PICTOGRAPHIC`?  I'm a little confused about how this rule corresponds to GB11.

utf8proc.c
```diff
@@ -299,8 +299,9 @@ static utf8proc_bool grapheme_break_extended(int lbc, int tbc, utf8proc_int32_t
     // second RI's bound class to UTF8PROC_BOUNDCLASS_OTHER, to force a break
     // after that character according to GB999 (unless of course such a break is
     // forbidden by a different rule such as GB9).
-    if (*state == tbc && tbc == UTF8PROC_BOUNDCLASS_REGIONAL_INDICATOR)
```
stevengj, line 302: Seems like this could just be:
```c
if (lbc_override== tbc && tbc == UTF8PROC_BOUNDCLASS_REGIONAL_INDICATOR)
```

utf8proc.c
kdheepak: I've replicated the logic that is `utfproc.c` in Julia. 

```julia
julia --project
               _
   _       _ _(_)_     |  Documentation: https://docs.julialang.org
  (_)     | (_) (_)    |
   _ _   _| |_  __ _   |  Type "?" for help, "]?" for Pkg help.
  | | | | | | |/ _` |  |
  | | |_| | | | (_| |  |  Version 1.5.2 (2020-09-23)
 _/ |\__'_|_|_|\__'_|  |  Official https://julialang.org/ release
|__/                   |

julia> const utf8proc_propval_t = Int16
Int16

julia> const utf8proc_uint16_t = UInt16
UInt16

julia> struct utf8proc_property_t
         category::utf8proc_propval_t
         combining_class::utf8proc_propval_t
         bidi_class::utf8proc_propval_t
         decomp_type::utf8proc_propval_t
         decomp_seqindex::utf8proc_uint16_t
         casefold_seqindex::utf8proc_uint16_t
         uppercase_seqindex::utf8proc_uint16_t
         lowercase_seqindex::utf8proc_uint16_t
         titlecase_seqindex::utf8proc_uint16_t
         comb_index::utf8proc_uint16_t
         flags::UInt8
         boundclass::UInt8
         # unsigned bidi_mirrored:1;
         # unsigned comp_exclusion:1;
         # unsigned ignorable:1;
         # unsigned control_boundary:1;
         # unsigned charwidth:2;
         # unsigned pad:2;
         # unsigned boundclass:8;
       end

julia> prop(c::Char) = ccall(:utf8proc_get_property, Ref{utf8proc_property_t}, (UInt32, ), c)
prop (generic function with 1 method)

julia> @enum utf8proc_boundclass_t begin
         UTF8PROC_BOUNDCLASS_START              =  0 # /**< Start */
         UTF8PROC_BOUNDCLASS_OTHER              =  1 # /**< Other */
         UTF8PROC_BOUNDCLASS_CR                 =  2 # /**< Cr */
         UTF8PROC_BOUNDCLASS_LF                 =  3 # /**< Lf */
         UTF8PROC_BOUNDCLASS_CONTROL            =  4 # /**< Control */
         UTF8PROC_BOUNDCLASS_EXTEND             =  5 # /**< Extend */
         UTF8PROC_BOUNDCLASS_L                  =  6 # /**< L */
         UTF8PROC_BOUNDCLASS_V                  =  7 # /**< V */
         UTF8PROC_BOUNDCLASS_T                  =  8 # /**< T */
         UTF8PROC_BOUNDCLASS_LV                 =  9 # /**< Lv */
         UTF8PROC_BOUNDCLASS_LVT                = 10 # /**< Lvt */
         UTF8PROC_BOUNDCLASS_REGIONAL_INDICATOR = 11 # /**< Regional indicator */
         UTF8PROC_BOUNDCLASS_SPACINGMARK        = 12 # /**< Spacingmark */
         UTF8PROC_BOUNDCLASS_PREPEND            = 13 # /**< Prepend */
         UTF8PROC_BOUNDCLASS_ZWJ                = 14 # /**< Zero Width Joiner */

         # the following are no longer used in Unicode 11, but we keep
         # the constants here for backward compatibility
         UTF8PROC_BOUNDCLASS_E_BASE             = 15 # /**< Emoji Base */
         UTF8PROC_BOUNDCLASS_E_MODIFIER         = 16 # /**< Emoji Modifier */
         UTF8PROC_BOUNDCLASS_GLUE_AFTER_ZWJ     = 17 # /**< Glue_After_ZWJ */
         UTF8PROC_BOUNDCLASS_E_BASE_GAZ         = 18 # /**< E_BASE + GLUE_AFTER_ZJW */

         # the Extended_Pictographic property is used in the Unicode 11
         # grapheme-boundary rules, so we store it in the boundclass field
         UTF8PROC_BOUNDCLASS_EXTENDED_PICTOGRAPHIC = 19
         UTF8PROC_BOUNDCLASS_E_ZWG = 20 # /* UTF8PROC_BOUNDCLASS_EXTENDED_PICTOGRAPHIC + ZWJ */
       end

julia> function boundclass(c::Char)
         r = utf8proc_boundclass_t(prop(c).boundclass)
         return r
       end
boundclass (generic function with 1 method)
```

If we take the following unicode string, can call the boundclass function on it, we get the following:

```julia
julia> unicode_s = "🤦🏼‍♂️"
"🤦🏼\u200d♂️"

julia> for c in unicode_s
       @show boundclass(c)
       end
boundclass(c) = UTF8PROC_BOUNDCLASS_EXTENDED_PICTOGRAPHIC
boundclass(c) = UTF8PROC_BOUNDCLASS_EXTEND
boundclass(c) = UTF8PROC_BOUNDCLASS_ZWJ
boundclass(c) = UTF8PROC_BOUNDCLASS_EXTENDED_PICTOGRAPHIC
boundclass(c) = UTF8PROC_BOUNDCLASS_EXTEND
```

I've not read the spec enough to know whether the implementation here is incorrect and if those codepoints should correspond to `EXTENDED_PICTOGRAPHIC` or `EXTEND`.

test/graphemetest.c
```diff
@@ -80,5 +80,51 @@ int main(int argc, char **argv)
...
+        }
+
+        state = 0;
+        for (int i = 0; i < 5; i++) {
+            utf8proc_int32_t c1 = family[i];
+            utf8proc_int32_t c2 = family[i+1];
+            results_family[i] = utf8proc_grapheme_break_stateful(c1, c2, &state);
+            check(results_family[i] == expected_family[i], "Incorrect grapheme break on initial extended + zwj emoji");
+        }
+    }
```
stevengj, line 127: Can we also check that `utf8proc_map(utf8, j, &g, UTF8PROC_CHARBOUND);` works for these cases?

Conversely, we should check that manually calling `utf8proc_grapheme_break_stateful` works with the previous `GraphemeBreakTest.txt` testcases in addition to `utf8proc_map`.

utf8proc.c
stevengj: The codepoints are correct, but why does the `state` not stay in `UTF8PROC_BOUNDCLASS_EXTENDED_PICTOGRAPHIC` since that is the first character?

kdheepak: This is because the first "state" is always `UTF8PROC_BOUNDCLASS_START`. And in this case, it is `UTF8PROC_BOUNDCLASS_START` instead of `UTF8PROC_BOUNDCLASS_EXTENDED_PICTOGRAPHIC`. It then goes to `UTF8PROC_BOUNDCLASS_EXTENDED` instead.

kdheepak: This is also perhaps why it works as expected for strings where an ascii character is prefixed with the emoji, and we only encounter this issue when the emoji is at the very start of the string.

stevengj: Shouldn't the state change to `UTF8PROC_BOUNDCLASS_EXTENDED_PICTOGRAPHIC` after the first (emoji) codepoint?

stevengj:
It seems like a simpler fix would be:
```c
static utf8proc_bool grapheme_break_extended(int lbc, int tbc, utf8proc_int32_t *state)
{
    int lbc_override;
    if (*state = UTF8PROC_BOUNDCLASS_START)
        *state = lbc;
    else
        lbc_override = *state;
...
```
Isn't that really the underlying bug, that `*state` isn't being initialized correctly?

kdheepak:
Wouldn't `lbc_override` be uninitialized if `*state == UTF8PROC_BOUNDCLASS_START`? Do you mean this instead?

```
    int lbc_override = lbc;
    if (*state == UTF8PROC_BOUNDCLASS_START)
        *state = lbc; 
    else
        lbc_override = *state;
```

Thanks for the feedback. My initial patch was based on a limited understanding of what is going on in utfproc. Feel free to suggest any recommendations you have. 

I'm testing this now to see if your fix works.

kdheepak:
I checked out the latest master from this repo and added the following patch

```
diff --git a/utf8proc.c b/utf8proc.c
index 6591976..00abc2c 100644
--- a/utf8proc.c
+++ b/utf8proc.c
@@ -290,8 +290,11 @@ static utf8proc_bool grapheme_break_simple(int lbc, int tbc) {
 
 static utf8proc_bool grapheme_break_extended(int lbc, int tbc, utf8proc_int32_t *state)
 {
-  int lbc_override = ((state && *state != UTF8PROC_BOUNDCLASS_START)
-                      ? *state : lbc);
+  int lbc_override = lbc;
+  if (*state == UTF8PROC_BOUNDCLASS_START)
+      *state = lbc;
+  else
+      lbc_override = *state;
   utf8proc_bool break_permitted = grapheme_break_simple(lbc_override, tbc);
   if (state) {
     // Special support for GB 12/13 made possible by GB999. After two RI
```

along with @archermarx's tests from this PR, and all the tests pass.

stevengj:
Yes, sorry:
```c
   int lbc_override;
    if (*state == UTF8PROC_BOUNDCLASS_START)
        *state = lbc_override = lbc; 
    else
        lbc_override = *state;
```

stevengj:
I updated the PR to use the new fix.

archermarx:
Excellent!

stevengj:
Whoops, I accidentally overwrite your master with my master, rather than with my branch.  

I re-opened as #205, with updated tests so that it also tests the `utf8proc_map` interface.  Will merge once tests are green.

archermarx:
No worries!

PR #205, opened 2020-11-23
Fix grapheme breaks on string-initial

### LINKED ISSUE #203: Grapheme breaks

stevengj:
Fixes #203. JuliaLang/julia#37680 will also be fixed once Julia updates utf8proc.

Replaces #204.


archermarx:
The consolidated tests look great! Thanks so much for helping this along!

PR #213, opened 2021-01-04
add utf8proc_iterate_graphemes

niblo:
Wondering if you're interested in this, or at least as a starting point. The tests pass, but I expect faults to be found if you take a look.


stevengj:
It's an interesting idea, but I wonder if it would be better to simply put an example of a loop over graphemes into the manual?

niblo:
I think it's a great idea, but only probably because that's what I need. It's convenient to have it tested and ready to use.

Issue #215, opened 2021-01-22
OSS-Fuzz integration


randy408:
[OSS-Fuzz](https://google.github.io/oss-fuzz/) can be used to automate fuzz testing for this project. Once integrated issues will be automatically filed on bugs.chromium.org when bugs are found and closed when they're fixed. 

I intend to integrate this project, which would involve 2-3 PRs.

Are the authors interested in OSS-Fuzz integration?

https://github.com/google/oss-fuzz/pull/5030

stevengj:
I'm open to this; what kind of changes to the utf8proc repo would be required?

randy408:
Integration consists of one or more fuzz targets and a build script that runs in an oss-fuzz environment (those could all go under `test/`).

Adding the fuzz target(s) to the CMakefile behind an option is also very useful for reproducing test cases, the rest of the integration happens in the oss-fuzz repository. I'm intending to make the PR's for this on both repos.

stevengj:
Sounds good!

PR #216, opened 2021-01-23
OSS-Fuzz initial integration

randy408:
Adds a fuzz target, build script for integration. The fuzz target is also added to the CMake build when `UTF8PROC_ENABLE_TESTING` is enabled (build tested on linux + windows), this allows for reproducing test cases outside of a  OSS-Fuzz Docker environment.

Will follow up with PR's once this is up and running.

#215


randy408:
This is ready for review.

test/ossfuzz.sh
```diff
@@ -0,0 +1,17 @@
...
+mkdir build
+cd build
+cmake .. -DUTF8PROC_ENABLE_TESTING=ON
+make -j$(nproc)
+
+
+$CC $CFLAGS -I$SRC/utf8proc \
+    $SRC/utf8proc/test/fuzzer.c \
+    -o $OUT/utf8proc_fuzzer \
+    $LIB_FUZZING_ENGINE $SRC/utf8proc/build/libutf8proc.a
```
stevengj, line 14: Shouldn't this be built by cmake, e.g. with a `-DLIB_FUZZING_ENGINE="$LIB_FUZZING_ENGINE" flag?

randy408: It will now link against `$LIB_FUZZING_ENGINE` when defined before falling back to the standalone driver.

CMakeLists.txt
```diff
@@ -98,4 +98,12 @@ if(UTF8PROC_ENABLE_TESTING)
   target_link_libraries(normtest utf8proc)
   add_test(utf8proc.testgraphemetest graphemetest data/GraphemeBreakTest.txt)
   add_test(utf8proc.testnormtest normtest data/NormalizationTest.txt)
+  
+  if(DEFINED ENV{LIB_FUZZING_ENGINE})
```
stevengj, line 102: I'd rather require the user to explicitly pass `-DLIB_FUZZING_ENGINE=...` than use an environment variable, since environment variables are too vulnerable to being set accidentally.   The `ossfuzz.sh` script can then pass `-DLIB_FUZZING_ENGINE="$LIB_FUZZING_ENGINE"` using the environment variable.

randy408: Updated to use a build option.

stevengj:
LGTM, thanks!

PR #219, opened 2021-02-04
OSS-Fuzz integration updates

randy408:
This PR improves code coverage, fixes the build (currently is worked around) and adds CIFuzz integration. [CIFuzz](https://google.github.io/oss-fuzz/getting-started/continuous-integration/#how-it-works) integration is basically a GitHub Action with a build test and a short fuzz test for PRs, without it merged bugs would only be detected the next day at best.


stevengj:
LGTM.

PR #222, opened 2021-03-29
buildsys: cmake install .pc file for pkg-config

leleliu008:
modified `libutf8proc.pc.in`, changed from `PREFIX` to `@PREFIX@`, changed from `VERSION` to `@VERSION@`, beacuse `@VAR@` can be replaced by CMake's `configure_file` command.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -51,15 +51,21 @@ set_target_properties (utf8proc PROPERTIES
 )
 
 if (UTF8PROC_INSTALL)
-  install(TARGETS utf8proc
-    RUNTIME DESTINATION bin
-    LIBRARY DESTINATION lib
-    ARCHIVE DESTINATION lib)
+  include(GNUInstallDirs)
 
-  install(
-    FILES
-      "${PROJECT_SOURCE_DIR}/utf8proc.h"
-    DESTINATION include)
+  install(TARGETS utf8proc
+    RUNTIME DESTINATION ${CMAKE_INSTALL_BINDIR}
+    LIBRARY DESTINATION ${CMAKE_INSTALL_LIBDIR}
+    ARCHIVE DESTINATION ${CMAKE_INSTALL_LIBDIR})
+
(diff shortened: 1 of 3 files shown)


version.txt
```diff
@@ -0,0 +1 @@
+2.6.1
```
stevengj, line 1: I'd rather not have this file — it's yet another place we'll have to update when we bump the version. Why can't the make variables be used?

leleliu008: OK. I will remove this file.

CMakeLists.txt
```diff
@@ -60,6 +60,13 @@ if (UTF8PROC_INSTALL)
     FILES
       "${PROJECT_SOURCE_DIR}/utf8proc.h"
     DESTINATION include)
+
+  set(PREFIX ${CMAKE_INSTALL_PREFIX})
+  set(LIBDIR lib)
+  set(INCLUDEDIR include)
+  set(VERSION 2.6.1)
+  configure_file(libutf8proc.pc.in  libutf8proc.pc @ONLY)
+  install(FILES ${CMAKE_BINARY_DIR}/libutf8proc.pc DESTINATION lib/pkgconfig)
```
tristan957, line 69: Hardcoded destination feels wrong. You should use GNUInstallDirs so downstream users don't have to patch `lib` to be `lib64`

tristan957: Kind of dependent on https://github.com/JuliaStrings/utf8proc/pull/159/files I guess.

leleliu008: Thank you for your suggestion. I have changed it.

diizzyy:
I think something like this would reduce maintence even further, not by a huge margin though.
It would need to use the Makefile instead of configure.ac
https://github.com/libcheck/check/blob/master/CMakeLists.txt#L61

CMakeLists.txt
stevengj: I've merged #159, so one can use that now.

stevengj:
closedd by #224

PR #223, opened 2021-04-18
Use stdint.h instead of inttypes.h

mwilliamson:
This improves support for targeting wasm32 with clang 12, and I don't think anything else in inttypes.h is used that isn't already in stdint.h.


stevengj:
Seems fine to me, thanks!

Issue #221, opened 2021-02-26
2.6.1: cmake do not generate ansd install libutf8proc.pc


kloczek:
Yep looks like that part is not finished.

leleliu008:
this is a duplicate with #198 , I can fix this, do you accept PR?

stevengj:
fixed by #224

Issue #200, opened 2020-10-09
[Patch] Install libraries in CMAKE_INSTALL_LIBDIR


epsilon-0:
Instead of always using `lib` use the default library location as defined by the system settings

The patch is from Gentoo Science project.

```
diff --git a/CMakeLists.txt b/CMakeLists.txt
index bfc3f9b..8fef4af 100644
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -52,8 +52,8 @@ set_target_properties (utf8proc PROPERTIES
 if (UTF8PROC_INSTALL)
   install(TARGETS utf8proc
     RUNTIME DESTINATION bin
-    LIBRARY DESTINATION lib
-    ARCHIVE DESTINATION lib)
+    LIBRARY DESTINATION ${CMAKE_INSTALL_LIBDIR}
+    ARCHIVE DESTINATION ${CMAKE_INSTALL_LIBDIR})
 
   install(
     FILES
```

Thanks,
Aisha

stevengj:
fixed by #224

Issue #198, opened 2020-08-23
cmake doesn't install .pc file for pkg-config


sternenseemann:
As far as I can see only the `make` install process generates and installs the pkg-config file. Would be great if that was also supported in the `cmake` installation process.

stevengj:
fixed by #224

PR #224, opened 2021-04-25
Fix cmake install

markus-oberhumer:


markus-oberhumer:
I've just pushed a new version - this should also include the functionality of #222 .

CMakeLists.txt
```diff
@@ -1,10 +1,13 @@
...
 disallow_intree_builds()
 
-project (utf8proc C)
+if (POLICY CMP0048)
+  cmake_policy (SET CMP0048 NEW)
+endif ()
+project (utf8proc VERSION 2.6.1 LANGUAGES C)
 
 # This is the ABI version number, which may differ from the
 # API version number (defined in utf8proc.h).
```
stevengj, line 13: Maybe add a comment here to remind us that the API version number should also be set in the `project` statement above when it is updated in utf8proc.h.

markus-oberhumer: > Maybe add a comment here to remind us that the API version number should also be set in the `project` statement above when it is updated in utf8proc.h.

Might be a good idea - please merge this and then adapt to your liking.

markus-oberhumer:
Please merge this - the current git is broken as it installs headers into a directory called `include64`.

Furthermore this would also fix the following issues: #198, #200, #221, #222

PR #229, opened 2021-10-09
[ci] set github CI

### LINKED ISSUE #206: migrate CI from Travis to github actions?
Travis queues are really slow these days, and the writing seems to be on the wall for free CI — we'll probably have no choice but to migrate eventually.

I haven't set up github actions CI yet myself, so if someone is already familiar with this it would be great to have a PR.

inkydragon:
Build on: ubuntu, windows, macOS, win+mingw64

Fix #206

We can add more combinations if needed, for example: ubuntu+mingw64, win+cygwin, *+clang.

Results of CI run: https://github.com/inkydragon/utf8proc/actions/runs/1324116342

### DIFF
--- a/.github/workflows/cmake.yml
+++ b/.github/workflows/cmake.yml
@@ -1,0 +1,64 @@
+name: CMake
+
+on:
+  push:
+    branches:
+      - master
+      - 'release-*'
+  pull_request:
+  # run on all pr
+
+jobs:
+  build:
+    strategy:
+      matrix:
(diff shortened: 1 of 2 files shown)


inkydragon:
> the Travis script

My understanding is that the `.travis.yml` mainly tests the usability of `make`.
I will create a new make.yml to test make.

The part that tests `cmake` is already included in the new `cmake.yml`.

PR #235, opened 2022-01-07
[CI] remove the MANIFEST file completely

inkydragon:
+ rm MANIFEST file
+ update CI yml


stevengj:
I would rather just remove the `MANIFEST` file completely; it's yet another thing to keep in sync, and it's not clear how useful it is...

inkydragon:
@stevengj MANIFEST is completely removed.

PR #240, opened 2022-06-08
Fix ASAN errors due adding offset to nullptr

parmeet:
This PR fixes the error when Address Sanitization is enabled. The offset is added to nullptr which gives following runtime error:
```
applying non-zero offset 4 to null pointer
```

The PR checks whether the pointer is nullptr, and if so do not add offset to it.

### DIFF
--- a/utf8proc.c
+++ b/utf8proc.c
@@ -365,9 +365,16 @@ static utf8proc_ssize_t seqindex_write_char_decomposed(utf8proc_uint16_t seqinde
   for (; len >= 0; entry++, len--) {
     utf8proc_int32_t entry_cp = seqindex_decode_entry(&entry);
 
-    written += utf8proc_decompose_char(entry_cp, dst+written,
-      (bufsize > written) ? (bufsize - written) : 0, options,
-    last_boundclass);
+    if (!dst){
+      written += utf8proc_decompose_char(entry_cp, dst,
+        (bufsize > written) ? (bufsize - written) : 0, options,
+      last_boundclass);
+    }else{
+      written += utf8proc_decompose_char(entry_cp, dst+written,
+        (bufsize > written) ? (bufsize - written) : 0, options,
+      last_boundclass);
+    }
+
     if (written < 0) return UTF8PROC_ERROR_OVERFLOW;
   }
   return written;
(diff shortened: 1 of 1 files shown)


utf8proc.c
```diff
@@ -365,9 +365,16 @@ static utf8proc_ssize_t seqindex_write_char_decomposed(utf8proc_uint16_t seqinde
...
+    if (!dst){
+      written += utf8proc_decompose_char(entry_cp, dst,
+        (bufsize > written) ? (bufsize - written) : 0, options,
+      last_boundclass);
+    }else{
+      written += utf8proc_decompose_char(entry_cp, dst+written,
+        (bufsize > written) ? (bufsize - written) : 0, options,
+      last_boundclass);
+    }
+
```
stevengj, line 377: Maybe cleaner to combine into a single call with a `dst ? ...` argument
```suggestion
    written += utf8proc_decompose_char(entry_cp, 
      dst ? dst+written : dst,
      (bufsize > written) ? (bufsize - written) : 0, options,
      last_boundclass);
```

stevengj:
(Note that the ASAN warning is harmless in this case, since `utf8proc_decompose_char` doesn't write anything if `bufsize` is 0 as will be the case here.  But we might as well silence ASAN here.)

parmeet:
@stevengj  This is my rough understanding what's going on:

When [utf8proc_decompose_custom](https://github.com/JuliaStrings/utf8proc/blob/63f31c908ef7656415f73d6c178f08181239f74c/utf8proc.c#L520) is called here https://github.com/JuliaStrings/utf8proc/blob/63f31c908ef7656415f73d6c178f08181239f74c/utf8proc.c#L736 inside utf8proc_map_custom with buffer == NULL, and wpos is incremented inside the while loop, it is adding wpos (>0) to buffer which is NULL. This give run-time error `applying non-zero offset 4 to null pointer`. By incorporating changes in this PR, I could get rid of these run-time errors.

stevengj:
That sounds like a runtime error caused by ASAN.  Applying a nonzero offset to a null pointer is perfectly safe if you never dereference the pointer.

xaizek:
> That sounds like a runtime error caused by ASAN. Applying a nonzero offset to a null pointer is perfectly safe if you never dereference the pointer.

Answers to [this question](https://stackoverflow.com/q/22103758/1535516) suggest that it's a UB. Although I'm not sure how come NULL pointer is "a pointer into, or just beyond, an array object", but clang generates an undefined behaviour error for this which suggest it might be a valid interpretation.

eschnett:
Memory access violations in C string processing libraries are a major source of bugs. Thus tools like ASAN and MSAN are very important. I think we should merge this PR and avoid all other similar C standard violations.

stevengj:
Fair enough.

PR #255, opened 2023-11-23
upgrade minimum cmake version

dundargoc:
This will silence the following warning:

CMake Deprecation Warning at CMakeLists.txt:1 (cmake_minimum_required):
  Compatibility with CMake < 3.5 will be removed from a future version of
  CMake.

  Update the VERSION argument <min> value or use a ...<max> suffix to tell
  CMake that the project does not need compatibility with older versions.


stevengj:
Sounds good, thanks.

PR #258, opened 2023-12-10
Port ruby data_generator.rb to Julia

c42f:
This reduces the number of dependencies needed when regenerating the C code by removing ruby from the tool chain. In addition, the bootstrapping requirement that libutf8proc needs to be built before CharWidths.txt can be generated is removed. The new code also separates C code generation from unicode data analysis more cleanly.

The output C code is identical to the original Ruby script, for now. Some bugs which were found in the process are noted as FIXMEs in the Julia source and can be fixed next.

This version of the script should also be better factored for generating a Julia version of the data file in the future (see https://github.com/JuliaLang/JuliaSyntax.jl/pull/381 for the somewhat unfortunate reasons that this seems like a good idea).


c42f:
Bump. Any interest in this?

data/data_generator.jl
```diff
@@ -0,0 +1,582 @@
...
+ignorable = Set(collect_codepoints(derived_core_properties, "Default_Ignorable_Code_Point"))
+uppercase = Set(collect_codepoints(derived_core_properties, "Uppercase"))
+lowercase = Set(collect_codepoints(derived_core_properties, "Lowercase"))
+
+
+#-------------------------------------------------------------------------------
+function derive_indic_conjunct_break(derived_core_properties)
+    props = Dict{UInt32, String}()
+    for i in collect_codepoints(derived_core_properties, "InCB; Linker")
+        props[i] = "LINKER"
```
stevengj, line 43: Can’t this pattern use `setindex!(Ref(props), codepoints, “VALUE”)` here and elsewhere?

data/data_generator.jl
```diff
@@ -0,0 +1,582 @@
...
+        comb_indices[dm1] = 0x8000 | (comb2nd_indices[dm1] + offset)
+        @assert comb2nd_indices[dm1] + offset <= 0x4000
+        if dm1 in comb2nd_indices_nonbasic
+            comb_indices[dm1] |= 0x4000
+            offset += 1
+        end
+    end
+end
+
+function utf16_encode(utf32_seq)
```
stevengj, line 344: Can’t Julia’s `transcode` function do this?

data/data_generator.jl
c42f: That works (adding the forgotten `.` to make it a broadcast?) but it seems kind of obscure to me. I'll add a utility function.

data/data_generator.jl
c42f: It seems like it can't, not directly.  But it seems we can use `transcode` if we go via UTF-8. I could do that if you prefer?

```julia
julia> utf32_seq = transcode(UInt32, "a🎏") # for example
2-element Vector{UInt32}:
 0x00000061
 0x0001f38f

julia> transcode(UInt16, utf32_seq)
ERROR: MethodError: no method matching transcode(::Type{UInt16}, ::Vector{UInt32})

Closest candidates are:
  transcode(::Type{UInt8}, ::Vector{<:Union{Int32, UInt32}})
   @ Base strings/cstring.jl:170
  transcode(::Type{UInt16}, ::AbstractVector{UInt8})
   @ Base strings/cstring.jl:181
  transcode(::Type{String}, ::Any)
   @ Base strings/cstring.jl:179
  ...

Stacktrace:
 [1] top-level scope
   @ REPL[91]:1

julia> transcode(UInt16, transcode(String, utf32_seq))
3-element Vector{UInt16}:
 0x0061
 0xd83c
 0xdf8f
```

stevengj: Yes, let's just use `transcode` here — even though it is slightly less efficient to go through utf-8, maximizing efficiency isn't a big concern.

data/Makefile
```diff
@@ -1,29 +1,27 @@
...
 
 .DELETE_ON_ERROR:
 
-utf8proc_data.c.new: data_generator.rb UnicodeData.txt GraphemeBreakProperty.txt DerivedCoreProperties.txt CompositionExclusions.txt CaseFolding.txt CharWidths.txt emoji-data.txt
-	$(RUBY) data_generator.rb < UnicodeData.txt > $@
-
-CharWidths.txt: charwidths.jl EastAsianWidth.txt
-	$(JULIA) charwidths.jl > $@
+utf8proc_data.c.new: data_generator.jl rawdata
+	$(JULIA) --project=. data_generator.jl > $@
```
inkydragon, line 18: We may need to add a new target to initialize the julia environment first.

```suggestion
utf8proc_data.c.new: data_generator.jl rawdata julia-pkg-instantiated
	$(JULIA) --project=. data_generator.jl > $@

julia-pkg-instantiated:
	$(JULIA) --project=. -e 'using Pkg; Pkg.instantiate()'
	echo 1 > $@
```

stevengj: Since we only need it for the generator, maybe just combine the rules:
```suggestion
utf8proc_data.c.new: data_generator.jl rawdata
	$(JULIA) --project=. -e 'using Pkg; Pkg.instantiate()'
	$(JULIA) --project=. data_generator.jl > $@
```

data/Makefile
```diff
@@ -1,29 +1,28 @@
...
-	$(JULIA) charwidths.jl > $@
+utf8proc_data.c.new: data_generator.jl rawdata
+	$(JULIA) --project=. -e 'using Pkg; Pkg.instantiate()'
+	$(JULIA) --project=. data_generator.jl > $@
 
 # Unicode data version (must also update utf8proc_unicode_version function)
 UNICODE_VERSION=15.1.0
 
+rawdata: UnicodeData.txt GraphemeBreakProperty.txt DerivedCoreProperties.txt CompositionExclusions.txt CaseFolding.txt EastAsianWidth.txt emoji-data.txt
+
```
stevengj, line 25:
```suggestion
RAWDATA = UnicodeData.txt GraphemeBreakProperty.txt DerivedCoreProperties.txt CompositionExclusions.txt CaseFolding.txt EastAsianWidth.txt emoji-data.txt

utf8proc_data.c.new: data_generator.jl $(RAWDATA)
	$(JULIA) --project=. -e 'using Pkg; Pkg.instantiate()'
	$(JULIA) --project=. data_generator.jl > $@

# Unicode data version (must also update utf8proc_unicode_version function)
UNICODE_VERSION=15.1.0

```

data/Makefile
```diff
@@ -1,25 +1,24 @@
...
 # programs
 CURL=curl
-RUBY=ruby
 PERL=perl
 MAKE=make
 JULIA=julia
 CURLFLAGS = --retry 5 --location
 
-.PHONY: clean
+.PHONY: clean rawdata
```
stevengj, line 13:
```suggestion
.PHONY: clean
```

data/Makefile
```diff
@@ -52,12 +51,12 @@ emoji-data.txt:
...
-	$(RUBY) -e 'puts File.read("DerivedCoreProperties.txt")[/# Derived Property: Uppercase.*?# Total code points:/m]' > $@
+	$(JULIA) -e 'print(match(r"# Derived Property: Uppercase.*?# Total code points:"s, read("DerivedCoreProperties.txt", String)).match)' > $@
 
 Lowercase.txt: DerivedCoreProperties.txt
-	$(RUBY) -e 'puts File.read("DerivedCoreProperties.txt")[/# Derived Property: Lowercase.*?# Total code points:/m]' > $@
+	$(JULIA) -e 'print(match(r"# Derived Property: Lowercase.*?# Total code points:"s, read("DerivedCoreProperties.txt", String)).match)' > $@
 
 clean:
-	rm -f UnicodeData.txt EastAsianWidth.txt GraphemeBreakProperty.txt DerivedCoreProperties.txt CompositionExclusions.txt CaseFolding.txt NormalizationTest.txt GraphemeBreakTest.txt CharWidths.txt emoji-data.txt
+	rm -f UnicodeData.txt EastAsianWidth.txt GraphemeBreakProperty.txt DerivedCoreProperties.txt CompositionExclusions.txt CaseFolding.txt NormalizationTest.txt GraphemeBreakTest.txt emoji-data.txt
```
stevengj, line 60:
```suggestion
	rm -f $(RAWDATA) NormalizationTest.txt GraphemeBreakTest.txt
```

data/Makefile
c42f: Oops! Thanks

c42f:
Nice, looks like we're good to go here.

data/data_generator.jl
```diff
@@ -0,0 +1,573 @@
...
+            @assert !(cp & 0b1111100000000000 == 0b1101100000000000)
+            push!(enc, cp)
+        else
+            temp = cp - 0x10000
+            push!(enc, (temp >> 10) | 0b1101100000000000)
+            push!(enc, (temp & 0b0000001111111111) | 0b1101110000000000)
+        end
+    end
+    enc
+end
```
stevengj, line 349:
```suggestion
utf16_encode(utf32_seq) = transcode(UInt16, transcode(String, utf32_seq))
```

data/data_generator.jl
c42f: Efficiency is no concern but I do prefer the explicit version because it mirrors the decoding on the C side in a much more obvious way.

stevengj: It's a lot clearer not to duplicate functionality from Base here.

stevengj:
Thanks for working on this.  Care to open up PRs for the FIXMEs?

c42f:
> Care to open up PRs for the FIXMEs?

Will do!

Issue #226, opened 2021-07-19
A couple of potential issues in data_generator.rb


chris0e3:
Hello,

Thanks for all your work on utf8proc.  It’s much appreciated.

While rewriting data_generator.rb & charwidths.jl (v2.6.1) in Python I spotted a couple of potential issues.

1. The following lines in data_generator.rb produce spurious 0s which are added to $exclusions and $excl_version.  (This occurs because there are comment lines in the input.)
```
    134: $exclusions = $exclusions.chomp.split("\n").collect { |e| e.hex }
    ...
    137: $excl_version = $excl_version.chomp.split("\n").collect { |e| e.hex }
```
This results in `utf8proc_property_struct.comp_exclusion = true` for U+0000.  Without the spurious 0s it is `false`.

2. The following line in data_generator.rb looks wrong:
```
    250:    "#{%W[Zl Zp Cc Cf].include?(category) and not [0x200C, 0x200D].include?(category)}, " <<
                                                                                    ^^^^^^^^
```
should (probably) be:
```
    250:    "#{%W[Zl Zp Cc Cf].include?(category) and not [0x200C, 0x200D].include?(code)}, " <<
                                                                                    ^^^^
```
This results in `utf8proc_property_struct.control_boundary = true` for U+200C and U+200D.  With the change it is `false`.

Can anyone definitively state if these property changes are correct?
I hope this info is helpful.

If there’s interest I will open a separate issue for my new Python data generator.

Regards,

CHRIS

stevengj:
cc @c42f

c42f:
Huh. Completely correct IMO. I discovered both these fixes independently.

Luckily the consequences of these bugs seem quite minor, as documented in #259

c42f:
I think this can be closed now :)

PR #259, opened 2024-01-04
Remove ruby compat hacks

c42f:
Follow on from #258 

There's three minor cleanups here arranged into two patches so you can see the effect on utf8proc_data.c. (We might want to squash the patches together when merging because they churn utf8proc_data.c a bit due to two of them affecting the encoding tables in a non-meaningful but global way which is why I've put them all together here in this PR.)

Only one of these fixes is observable via the API as a minor bug fix for properties of U+200C and U+200D.

The patch notes:

### [PATCH 1/2] Fix two minor bugs from the Ruby code

First, `categroy` rather than `code` was used in constructing the
`control_boundary` property as related to the characters U+200C and
U+200D. This seemed incorrect and should be fixed. This could be an
observable bugfix for any C code which inspects the `control_boundary`
property.

Second, when reading composition exclusions, Ruby's String hex method
produces zero rather than nil if no number is found. For example

    $ ruby -e 'puts "# blah".hex'
    0

This led to the character `'\0'` being included in the `exclusions`
and `excl_versions` sets which is incorrect. However this seems
asymptomatic because `'\0'` is never part of a composition. (In terms of
the C code, the use of `comp_exclusion` is guarded by the `comb_index`
property which is `UINT16_MAX` for `'\0'`.)

### [PATCH 2/2] Cleanup: Remove sequence ordering hack

This hack changed the ordering of sequences encoded in the sequences
table and was added so we could easily prove equivalence to the Ruby
data generator code.

However, it's no longer needed and removing it shouldn't result in any
functional change.


stevengj:
Also fixes #226?

stevengj:
> This hack changed the ordering of sequences encoded in the sequences table and was added so we could easily prove equivalence to the Ruby data generator code.

Is the sequence ordering now deterministic independent of the Julia version?  I wouldn't want the output to change just because Julia changed its `Dict` iteration order.

c42f:
> Is the sequence ordering still deterministic independent of the Julia version? I wouldn't want the output to change just because Julia changed its Dict iteration order.

Yes I've kept iteration order separate from `Dict` - it's consistent with ruby which iterates hash in the order of insertion. (In the port one extra array `comb2nd_indices_sorted_keys` was introduced to ensure this.)

The only ordering-related thing which changes here is that we insert the decomposition mapping and case folding sequences into the cache interleaved with uppercase, lowercase and titlecase mappings, rather than doing them first. But that's independent of Julia version. The ordering in the Ruby code was less obvious because the mutable cache of sequences is a global variable.

PR #262, opened 2024-04-07
build: remove unnecessary policy check

dundargoc:
Minimum version is 3.5 and policy CMP0048 was introduced in 3.0, meaning
that it will always be set to `NEW`.


stevengj:
LGTM, thanks.

PR #268, opened 2024-07-10
ci(macos): install julia dependency

dundargoc:
Otherwise the job fails with the error message

"/bin/sh: julia: command not found"


stevengj:
Thanks!

PR #276, opened 2024-12-15
Add CODEOWNERS file to autoamtically add reviewers

dundargoc:


dundargoc:
@stevengj unsure if there are other maintainers, if so let me know and I can add them as well.

PR #289, opened 2025-03-16
fix improper integer type on systems with non-32-bit int

mrpapersonic:
generally the parameter here should be `utf8proc_int32_t` anyway, but changing public facing APIs for something like this isn't a very good idea. maybe this should be changed for the next major bump in the libtool version


stevengj:
`seqindex_write_char_decomposed` is not part of the API, so we can just change the parameter.

stevengj:
I'd rather just change the internal API and not the external ABI.

PR #326, opened 2026-01-03
Add missing CMake library alias

HexDecimal:
Without this, `target_link_libraries(${PROJECT_NAME} PRIVATE utf8proc::utf8proc)` only works with `find_package` and not with `add_subdirectory`. This is needed for `FetchContent` to work optimally.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -22,6 +22,7 @@ add_library (utf8proc
   utf8proc.c
   utf8proc.h
 )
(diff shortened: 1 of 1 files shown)


stevengj (approved):

stevengj:
Seems fine to me.

Wirtos:
This is approved long ago but not pushed? 
Created a similar PR myself but just noticed this one already exists and was approved.

PR #342, opened 2026-06-09
Add Bazel build system support

callframe:
I often use utf8proc via Bazel and found I had to repeatedly inject my own Bazel files into the cloned repository. This PR adds Bazel support upstream in case others would benefit too.

### What's included

The build currently produces only the library, since that's what consumers most commonly need. Thanks to Bazel's flexibility, extending support to the test files, benchmarks, and packaging should be straightforward if there's interest.

### Usage

Builds can be run from the command line:

```
bazel build //:utf8proc
```
Or inside another Bazel project, by adding the following to your `MODULE.bazel`:

```starlark
bazel_dep(name = "utf8proc")  # add version = "x.y.z" to pin a specific version
git_override(
    module_name = "utf8proc",
    remote = "https://github.com/callframe/utf8proc.git",  # or https://github.com/JuliaStrings/utf8proc once merged
    # pin to a commit for reproducible builds; branch shown here for convenience
    branch = "master",
)
```

The `git_override` is only needed because the module isn't published yet. If utf8proc were added to the [Bazel Central Registry](https://registry.bazel.build), the `bazel_dep` line alone would be enough and dependents could pull it directly from the registry.

### Notes

- Bazel 7.0.0 and newer are supported.
- The `WORKSPACE` mechanism is not supported, as it was removed in Bazel 9.0.0.
- The changes are minimally invasive: no preexisting files are modified except `.gitignore` and `README.md`.
- A Bazel lockfile is included. It pins the transitive dependency graph for reproducible builds, but it's optional and can be removed from version control if the project prefers not to track it.

Let me know how you feel about this, or whether there's interest in Bazel support at all.


stevengj:
I'm not sure we want to maintain yet another build system in this repository… every one we add comes with a maintenance cost.

(Note that for people who don't want to use utf8proc's included build, or simply link it as a library that is installed separately, it can be included in any project as literally a single file that has to be compiled and linked in, + includes.)

PR #345, opened 2026-06-23
draft unicode 18 support

### LINKED ISSUE #344: RFE: Unicode 18.0 support
Title says it all - a request for Unicode 18.0 support. The new version of the standard is due out in September. Any plans to support it?

Currently, `graphemetest` fails on some 16 odd tests with the 18.0 version of `GraphemeBreakTest.txt`. (This is after rebuilding utf8proc_data.c by changing the Unicode version in data/Makefile to 18)

 For example,

```
D:\src\utf8proc\build\Debug>graphemetest.exe GraphemeBreakTest.txt
checking line 100...
FAILED at line 179: grapheme mismatch: "/αÑì/αñò" instead of "/αÑìαñò"
```

And BTW, thanks much for this library which we use in the Tcl 9 language core.

stevengj:
Will close #344, but do not merge until the final Unicode 18 is released.

Other than new data tables, we needed to update our grapheme-break rules since GB9c has changed.  (Actually, the rule has gotten significantly simpler, which is a welcome change from the ever more-complicated segmentation rules that we typically see.)


stevengj:
For reference, here is the rationale for the simplified GB9c rule from the [L2/26-096 meeting notes](https://www.unicode.org/L2/L2026/26096-pag-report-utc187.pdf):

image

clason:
Superseded by #354?

stevengj:
Oh, I totally forgot about this PR when reviewing PR #354.

@eschnett, the implementation of GB9c in this PR seems simpler than what was merged, so I'll probably open a new PR with this version.

PR #347, opened 2026-08-19
Update CMakeLists.txt to fix link warning LNK4098.

krishna116:
[English]

I encountered this issue while using the statically compiled utf8proc (version 2.11.3): LINK: warning LNK4098: The default library “MSVCRT” conflicts with other libraries; please use /NODEFAULTLIB:library

Here is the solution provided by AI:
When you encounter the message “The default library ‘MSVCRT’ conflicts with other libraries” while using a static C++ library, it is usually because the static library is linked to a C runtime library (CRT) that is incompatible with your project. To resolve this issue, you need to ensure that the static library does not statically link the CRT, but instead depends on the version of the CRT used by your project.

[Chinese]

我使用静态编译的utf8proc(version 2.11.3)遇到这个问题：LINK : warning LNK4098: 默认库“MSVCRT”与其他库的使用冲突；请使用 /NODEFAULTLIB:library

以下是AI给出的解决方案：
在使用静态C++库时遇到“默认库‘MSVCRT’与其他库的使用冲突”提示，通常是因为该静态库链接了与您的项目不兼容的C运行时库（CRT）。要解决这个问题，您需要确保静态库不静态链接CRT，而是依赖于您的项目所使用的CRT版本。


stevengj:
This seems like it potentially could simply make some errors silent, allowing it to compile and link but crashing at runtime, if utf8proc is compiled with `/MT` and the user's code is compiled with `/MD` —  https://learn.microsoft.com/en-us/cpp/c-runtime-library/potential-errors-passing-crt-objects-across-dll-boundaries?view=msvc-170 … though the documentation is not very clear to me on how `NODEFAULTLIB` interacts with all of this.

(If `NODEFAULTLIB` were completely safe, why wouldn't it be the default when linking static libraries?)

If you want to link utf8proc statically and need it to use your C runtime, I would just directly compile `utf8proc.c` as part of your program.  It's literally just one file (plus a couple `#include`d files); you hardly need a library for static linking.

stevengj:
@eschnett or @giordano, do you happen to understand the microsoft linking semantics here?

giordano:
No, sorry, I never used MSVC.

krishna116:
In my experience, the “/MD” parameter is the default option for MSVC. Since this CMakeLists.txt file does not specify a CRT type, it is unlikely to generate a “/MT” static library. 

Finally, I set `CMAKE_EXPORT_COMPILE_COMMANDS` and used the Ninja generator to export the compilation commands, which revealed the source of the problem: I had installed both MSVC (for compiling the code) and Clang (for formatting the code), and CMake had defaulted to Clang as the compiler, using the “/MT” compilation parameter. Below are all the exported compilation parameters:

  "command": "E:\\LLVM\\bin\\clang.exe -DUTF8PROC_EXPORTS -DUTF8PROC_STATIC -ID:/temp/test_utf8proc/utf8proc-2.11.3 -O0 -g -Xclang -gcodeview -D_DEBUG -D_DLL -D_MT -Xclang --dependent-lib=msvcrtd -O2 -std=c99 -pedantic -Wall -o CMakeFiles\\utf8proc.dir\\utf8proc.c.obj -c D:\\temp\\test_utf8proc\\utf8proc-2.11.3\\utf8proc.c",
  "file": "D:/temp/test_utf8proc/utf8proc-2.11.3/utf8proc.c",

If use the msvc compiler, it will generate a compilation command with the “/MD” parameter:

  "command": "C:\\PROGRA~1\\MICROS~1\\2022\\COMMUN~1\\VC\\Tools\\MSVC\\1439~1.335\\bin\\Hostx64\\x64\\cl.exe  /nologo -DUTF8PROC_EXPORTS -DUTF8PROC_STATIC -ID:\\temp\\test_utf8proc\\utf8proc-2.11.3 /DWIN32 /D_WINDOWS /W3 /MDd /Zi /Ob0 /Od /RTC1 /FoCMakeFiles\\utf8proc.dir\\utf8proc.c.obj /FdCMakeFiles\\utf8proc.dir\\utf8proc.pdb /FS -c D:\\temp\\test_utf8proc\\utf8proc-2.11.3\\utf8proc.c",

This might be a cmake bug(to select "/MT" as default compile parameter), and this issue can be closed.

Thanks.

krishna116:
During a conversation with an AI today, I happened to receive a prompt:
```
1. Use clang-cl.exe instead of clang.exe
On Windows, Clang has two main frontends:
clang.exe: For Unix-style command lines (MinGW/MSYS2 environments).
clang-cl.exe: For MSVC compatibility mode. It emulates the behavior of cl.exe and accepts MSVC flags such as /MD and /O2.
For most Windows projects, you should use clang-cl.exe, as it offers better compatibility with existing Windows build systems and libraries.

2. MSVC Runtime Library (CRT) Issues
Clang-CL uses MSVC’s dynamic CRT by default (/MD or /MDd).
If your project mixes third-party libraries compiled with MSVC and code compiled with Clang, you must ensure they use the same type of CRT (either both dynamically linked or both statically linked); otherwise, an LNK2038 error will occur.
```

LB--:
CMake 3.15 adds the [`MSVC_RUNTIME_LIBRARY`](https://cmake.org/cmake/help/v3.15/prop_tgt/MSVC_RUNTIME_LIBRARY.html) target property to handle this, by default initialized by the [`CMAKE_MSVC_RUNTIME_LIBRARY`](https://cmake.org/cmake/help/v3.15/variable/CMAKE_MSVC_RUNTIME_LIBRARY.html#variable:CMAKE_MSVC_RUNTIME_LIBRARY) variable, which users can set on the command line when building with CMake. I don't think the CMake script for this project needs any changes to accommodate that, in theory it should just work for people building with CMake 3.15 or newer. Maybe [`CMP0091`](https://cmake.org/cmake/help/v3.15/policy/CMP0091.html#policy:CMP0091) needs to be enabled though?

PR #351, opened 2026-08-30
Add minimal Zig build scripts

### LINKED ISSUE #350: Add Zig build script
[Zig](https://ziglang.org) is a programming language with strong C interoperability, making it particularly easy to include C dependencies in Zig projects (and vice versa) -- in fact, Zig can compile C code (using either a native bootstrap or LLVM), and the `build.zig` scripts can serve as a (frankly saner) [alternative to CMake even for pure C projects](https://github.com/neovim/neovim/blob/master/build.zig). Zig also comes with built-in cross-compilation support, including to WASM.

All this being said, would you consider distributing Zig build scripts (two files only: script and manifest) for `utf8proc`? If so, I can make a PR upstreaming the [ones used by Neovim](https://github.com/neovim/neovim/tree/master/deps/utf8proc). I understand that taking on maintenance for additional tooling -- that you don't use yourself -- is not very appealing, especially since Zig is still heavily in development. At least I can promise that these files won't bitrot easily since they'd be exercised by Neovim (and covered by CI). But if you deem this out of scope, feel free to close this issue and no hard feelings!

clason:
Targets Zig 0.16 (but works with Zig 0.17/master).

Compiles and installs a static library and headers for downstream
projects (Zig or C) using `build.zig` build scripts.

Closes #350 (I thought it easier to directly show what it would look like. Feel free to close the PR if you don't want these.)

Test with downloading the appropriate 0.16 tarball from https://ziglang.org/download/ and running `path-to-extracted-folder/zig build` at the repo root.


build.zig.zon
```diff
@@ -0,0 +1,14 @@
+.{
+    .name = .utf8proc,
+    .fingerprint = 0xdcddf08dbb8954c4,
+    .version = "2.11.3",
```
clason, line 4: This number would need to be incremented when making a new release, so this does come with a slight maintenance burden.

(The `fingerprint` is a static ID and does not need to be changed.)

clason:
@stevengj Is there any interest in this? (I don't mean to be pushy; I just saw the recent update to Unicode 18.0 and assume a new release will happen soon. Having this be part of the release would be quite helpful for Neovim.)

build.zig.zon
eschnett: We have version numbers in at least 5 different places already. This new spot doesn't matter much.

eschnett:
Question: If there is a build script here, would it make sense to at least smoke-test it in CI? I am thinking of a new CI stage that just runs this build script, maybe with minimal tests (expected output files are there).

clason:
Right now, the script only builds the shared lib, nothing else (as this is all we needed in Neovim). Converting the tests would be quite a bit more work, and I'm not sure how useful that would be? Or what kind of minimal test did you have in mind?

(A workflow just building the lib -- possibly in different modes -- would be pretty easy to add.)

eschnett:
I am looking for a test that catches syntax errors and other egregious problems in the build script. Just using it to build, and then checking that the shared library was actually built with the expected name, is good enough.

build.zig.zon
eschnett: The current version of the package is 2.12.0.

eschnett (changes requested):

build.zig.zon
clason: Right, forgot to update that.

clason: (Well, it _was_ 2.11.3 when I made the PR ;))

PR #352, opened 2026-09-06
normalize_utf32: skip out-of-range codepoints in the compose loop (fixes #288)

### LINKED ISSUE #288: Segmentation fault in utf8proc_normalize_utf32 with invalid options
### Description
While fuzzing utf8proc with AFL++, we discovered that pass invalid option flags to `utf8proc_normalize_utf32`, the API may trigger a segmentation fault. 

```
// Valid option test - works correctly
utf8proc_option_t valid_options = 
    UTF8PROC_STRIPCC | 
    UTF8PROC_NLF2LS | 
    UTF8PROC_NLF2PS;
utf8proc_normalize_utf32(copy, num_codepoints, valid_options);

// Invalid option test - triggers crash
utf8proc_option_t invalid_options = 0xFFFF;
utf8proc_normalize_utf32(copy, num_codepoints, invalid_options);
```

### Environment
- OS: Ubuntu 22.04
- Compiler: Clang 16.0.6
- AFL++ Version: afl-fuzz++4.22a

### Reproduction
```bash
afl-clang-fast -o fuzzer_afl fuzzer_afl.c \
    -I/path/to/utf8proc/include \
    -L/path/to/utf8proc/lib \
    -lutf8proc \
    -fsanitize=address

./fuzzer_afl 

```

### ASAN Log
```bash
AddressSanitizer:DEADLYSIGNAL
=================================================================
==3853974==ERROR: AddressSanitizer: SEGV on unknown address 0x558dccc4294c (pc 0x558dcbbee9c7 bp 0x00007f000617 sp 0x7ffc0d45aeb0 T0)
==3853974==The signal is caused by a READ memory access.
    #0 0x558dcbbee9c7 in unsafe_get_property /srv/scratch/PAG/ksx/Fuzz/Library/utf8proc/utf8proc.c:237:7
    #1 0x558dcbbee9c7 in utf8proc_normalize_utf32 /srv/scratch/PAG/ksx/Fuzz/Library/utf8proc/utf8proc.c:655:53
    #2 0x558dcbbea291 in main /srv/scratch/PAG/ksx/Fuzz/Library/utf8proc/Fuzz/reproduce/fuzzer_afl.c:79:13
    #3 0x7fb067499249 in __libc_start_call_main csu/../sysdeps/nptl/libc_start_call_main.h:58:16
    #4 0x7fb067499304 in __libc_start_main csu/../csu/libc-start.c:360:3
    #5 0x558dcbb2a3e0 in _start (/srv/scratch/PAG/ksx/Fuzz/Library/utf8proc/Fuzz/reproduce/fuzzer_afl+0x223e0) (BuildId: e62e3f9342618db53c46bbabc796a818d12e19b0)
```

### Fuzz file and input

[fuzzer_afl.c.zip](https://github.com/user-attachments/files/18929456/fuzzer_afl.c.zip)

[crash_input.zip](https://github.com/user-attachments/files/18929457/crash_input.zip)

The `utf8proc_reencode` function may have similar crash behavior with invalid options

AetherAI3:
Fixes #288.

Small one. `utf8proc_normalize_utf32`'s `UTF8PROC_COMPOSE` branch reads each `buffer[rpos]` and passes it straight to `unsafe_get_property`, guarded only by `if (current_char < 0) continue;` (that skip is meant for the grapheme-break sentinel written by `utf8proc_decompose_char`, not for general invalid input).

`unsafe_get_property` indexes `utf8proc_stage1table[uc >> 8]`, and `utf8proc_stage1table` is a `const utf8proc_uint16_t[4352]`. So a UTF-32 codepoint `>= 0x110000` in the input buffer produces `uc >> 8 >= 4352` and reads past the table.

The public wrapper `utf8proc_get_property` already handles both bounds with the pattern `uc < 0 || uc >= 0x110000 ? utf8proc_properties : unsafe_get_property(uc)`, and `utf8proc_decompose_char` uses `if (uc < 0 || uc >= 0x110000) return UTF8PROC_ERROR_NOTASSIGNED;`. The fix mirrors that same guard in the one call site that was missing it.

### Verification

Built master `0075ed7d` with clang + `-fsanitize=address,undefined`.

Minimal reproducer (buffer with one out-of-range codepoint):

```c
utf8proc_int32_t buf[] = { 0x0041, 0x0301, 0x110000 };
utf8proc_normalize_utf32(buf, 3, UTF8PROC_COMPOSE);
```

- Baseline (RED): UBSan says `utf8proc.c:237:7: runtime error: index 4352 out of bounds for type 'const utf8proc_uint16_t[4352]'`. ASan says `AddressSanitizer: global-buffer-overflow ... READ of size 2` inside `unsafe_get_property`, called from `utf8proc_normalize_utf32` line 669, `0 bytes after global variable 'utf8proc_stage1table' ... of size 8704`. Same failure class as the SEGV in #288.
- Post-fix (GREEN): same reproducer returns 1, no ASan/UBSan finding. The invalid codepoint is skipped in the compose pass, same treatment the loop already gives the grapheme-break sentinel above it.

Regression under ASan/UBSan (only the tests that don't need downloaded Unicode data — the sandbox this ran in couldn't reach unicode.org):

- `test/iterate` — 673 tests passed
- `test/valid` — SUCCEEDED
- `test/case` — 2942 tests SUCCEEDED
- `test/misc` — NFC round-trip SUCCEEDED, API version 17.0.0
- `test/charwidth` — 156996 chars SUCCEEDED
- `test/custom` — map_custom SUCCEEDED
- `test/printproperty` — OK

Hand-checked smoke: compose of `c a f e ́` returns `c a f é` (U+0063 U+0061 U+0066 U+00E9) unchanged. `utf8proc_reencode` of `hello 世界` still produces the correct UTF-8 bytes.

### Notes on scope

The reporter mentions `utf8proc_reencode` may crash the same way. `utf8proc_reencode` calls `utf8proc_normalize_utf32` first, so this fix covers it transitively. `utf8proc_encode_char` and `charbound_encode_char` (called by `utf8proc_reencode` after normalize) already reject `uc >= 0x110000` by returning 0.

Other `unsafe_get_property` callers were reviewed:

- `utf8proc_decompose_char` (`utf8proc.c:457`) — guarded at line 456 (`if (uc < 0 || uc >= 0x110000) return UTF8PROC_ERROR_NOTASSIGNED;`).
- `utf8proc_decompose_custom` canonical-reordering loop (`utf8proc.c:609-610`) — the buffer contents there come from `utf8proc_decompose_char`, which only writes validated codepoints, so those `unsafe_get_property` calls are safe.
- The second `unsafe_get_property(*starter)` in the compose loop (`utf8proc.c:698`) — `starter` points into a buffer position we just wrote a validated codepoint into, so it inherits the guard added above.

### Diff

+3/-2 in one file, one function.

### Base

Applied on top of master `0075ed7d0adba45682ee6bf7a83b10f8fd110163`.

### Credit

Reported by @shuangxiangkan (#288) with an AFL++ reproducer and ASan trace. Their trace pins the crash exactly at `utf8proc.c:237` — the finding did the analytical work.

### Note on AI assistance

I used an AI assistant to help trace the crash back to the missing bound and to draft this write-up. The code change and every claim above I verified myself against the built binary.


stevengj (dismissed):

stevengj:
This seems reasonable to me.  (Mostly we haven't used this function directly ourselves, since we only do normalization on UTF-8 text in Julia, so by the time it is decomposed to UTF-32 it is already validated.)

Can you add a test?  Just something in `test/misc.c` should be fine.

stevengj:
There was also a warning in the docstring that should now be removed:
```
 * @warning The entries of the array pointed to by `str` have to be in the
 *          range `0x0000` to `0x10FFFF`. Otherwise, the program might crash!
 ```

**Comment by eschnett** [comment]
I used an AI assistant to review this. It agrees with your PR and suggests correcting an additional bug. The case `current_char < 0` drops the `0xFF` marker instead of passing it through. Changing the line you modify to

```
if (current_char < 0) {
  /* grapheme-break sentinel: pass through and reset composition state */
  buffer[wpos++] = current_char;
  starter = NULL;
  starter_property = NULL;
  max_combining_class = -1;
  continue;
}
if (current_char >= 0x110000) {
  /* out-of-range codepoint: drop it (unsafe_get_property would OOB) */
  continue;
}
```

would correct this as well. This problem occurs only for `UTF8PROC_CHARBOUND | UTF8PROC_COMPOSE`, and we apparently do not have a test for this.

eschnett:
I opened https://github.com/JuliaStrings/utf8proc/pull/355 for the problem I mention above.

stevengj:
I merged this and pushed a separate commit c5afd53a40376d25939d07796ba019c261d4fae9 updating the docstring and the NEWS.  According to #355 it sounds like fuzz tests already catch this.

eschnett:
No, they do not. The fuzz tests test this code path, but they only look for crashes, they do not check whether the result is correct. The additional test I added does that.

PR #353, opened 2026-09-10
change options argument type to unsigned int

stevengj:
This PR changes argument-type declarations of the form `utf8proc_option_t options` to `unsigned int options`.  The former is technically incorrect (though it worked in practice) because this argument is *not* in general an enumerated value, but is rather a bitwise-OR of enumerated values.

(This change doesn't break the ABI, because the byte representation of `enum` is the same as that of `unsigned int` on all widely available systems, I think, even though strictly speaking this is not required by the C standard?)

I noticed this because gcc was emitting warnings like:
```
test/custom.c:20:99: warning: implicit conversion from
      'int' to enumeration type 'utf8proc_option_t' is invalid in C++
      [-Wimplicit-int-enum-cast]
   20 |     utf8proc_map_custom(input, 0, &output, UTF8PROC_CASEFOLD | UTF8PROC_COMPOSE | UTF8PROC_COMPAT | UTF8PROC_NULLTERM,
```

Originally, these option constants were `#define` preprocessor constants, but they were changed to an enum by #29.


eschnett:
After your change there are still casts to `(utf8proc_option_t)` in the code when doing bit manipulations. I think they can be removed.

Technically, this is an API change that deserves a NEWS entry.

stevengj:
Added NEWS entry and removed the obsolete typecasts.

PR #323, opened 2025-12-21
Correct OOB read

eschnett:
Closes https://github.com/JuliaStrings/utf8proc/issues/322.

Issue #144, opened 2018-11-16
Question about unsafe_encode_char()


ideal:
```c
   } else if (uc == 0xFFFF) {
       dst[0] = (utf8proc_uint8_t)0xFF;
       return 1;
   } else if (uc == 0xFFFE) {
       dst[0] = (utf8proc_uint8_t)0xFE;
       return 1;
   }
```

Though U+FFFE and U+FFFF are both not characters (https://www.unicode.org/charts/PDF/UFFF0.pdf), why are they treated specially encoding to one byte in unsafe_encode_char()? Thanks.

stevengj:
See #34.  These are ["noncharacters"](http://www.unicode.org/faq/private_use.html#nonchar1) and but we are required to encode them: *An implementation which converts noncharacter code points between one UTF representation and another must preserve these values correctly. The fact that they are called "noncharacters" and are not intended for open interchange does not mean that they are somehow illegal or invalid code points which make strings containing them invalid.*

However, the encoding here seems wrong to me — U+FFFF is supposed to encode to EFBFBF and U+FFFE is supposed to encode to EFBFBE, I thought.  I will look into it.

cc @ScottPJones, who was the last person to touch this code in #35.

stevengj:
Actually, `utf8proc_encode_char` is correct for these codepoints.  It's only the internal `unsafe_encode_char` that is doing something odd, I forget why…

stevengj:
Oh, that's right, we discussed this in #35.  The `CHARBOUND` feature internally uses 0xffff to insert grapheme markers as 0xff bytes.  But then this will get confused if there are actual 0xffff noncharacters in the string.

Nowadays we tend to recommend `utf8proc_grapheme_break_stateful` to find grapheme boundaries; we've talked about deprecating `CHARBOUND` for a while but haven't gotten around to it.

stevengj:
I think I can fix it to not screw up `U+FFF*` characters by using `0xffffffff` as the internal sentinel value instead.  As long as we have the `CHARBOUND` option, it is best to fix it to not screw up noncharacters.

ideal:
OK, thanks.

PR #355, opened 2026-09-21
normalize_utf32: keep grapheme-break sentinels in the compose loop

eschnett:
**Stacked on #352** — the base branch `pr352-base` is master + #352's commit, so the diff here shows only this change. Once #352 merges, I'll retarget this at `master` and delete `pr352-base`.

## The bug

`UTF8PROC_CHARBOUND` combined with `UTF8PROC_COMPOSE` or `UTF8PROC_DECOMPOSE` has emitted **no `0xff` grapheme markers at all** since v2.11.3.

#323 fixed a real out-of-bounds read by adding `if (current_char < 0) continue;` to the compose loop in `utf8proc_normalize_utf32`, but that `continue` *drops* the sentinel instead of passing it through — `wpos` is never advanced, so the marker disappears from the output.

```
                          pre-#323            v2.11.3 .. HEAD
"ab"    CHARBOUND         ff 61 ff 62         ff 61 ff 62
"ab"    CHARBOUND|NFC     ff 61 ff 62            61 62          <-- markers gone
"éxy"   CHARBOUND|NFC  ff c3 a9 ff 78 ff 79      c3 a9 78 79
```

Note this also reaches `UTF8PROC_DECOMPOSE`: `utf8proc_map` always routes through `utf8proc_reencode` → `utf8proc_normalize_utf32`, and the `COMPOSE` branch of that loop is where the sentinels are consumed.

## The fix

Write the sentinel to the output and reset the composition state at it, so composition never runs across a grapheme boundary. This restores the pre-#323 output byte-for-byte on every case I checked, while keeping #323's and #352's fix intact — the sentinel is still never passed to `unsafe_get_property`.

The out-of-range check #352 added moves to its own branch, because that case *does* want to drop the codepoint: `utf8proc_encode_char` already returns 0 for codepoints `>= 0x110000`, so dropping keeps the compose path consistent with the non-compose path right next to it.

## Why this wasn't caught

`CHARBOUND | COMPOSE` appeared in the test suite only in `test/fuzzer.c:96`, which doesn't check output. There was no correctness test for the combination, so #323 shipped the regression and `make check` stayed green.

This PR adds one next to the existing #144 `CHARBOUND` regression test in `test/graphemetest.c`, covering both `COMPOSE` and `DECOMPOSE`. Verified red/green — built against #352's head, the new test fails:

```
FAILED at line 883: CHARBOUND|COMPOSE returned length 4, expected 7
```

and passes with this commit.

## Verification

- `make check` passes.
- 400k-iteration random fuzz of `utf8proc_normalize_utf32` / `utf8proc_reencode` (in-range, out-of-range, negative and full-width random `int32_t` buffers across five option combos) under `-fsanitize=address,undefined -fno-sanitize-recover=all`: clean. Same harness trips `utf8proc.c:237: index 4352 out of bounds` immediately on master, and is clean on #352 — so this change doesn't reintroduce the OOB.
- Output compared against a pre-#323 build (`2cf0bf5^`) on the cases in the table above: identical.


utf8proc.c
```diff
@@ -662,9 +662,18 @@ UTF8PROC_DLLEXPORT utf8proc_ssize_t utf8proc_normalize_utf32(utf8proc_int32_t *b
...
+           (unsafe_get_property would OOB on utf8proc_stage1table) */
+        buffer[wpos++] = current_char;
+        starter = NULL;
+        starter_property = NULL;
+        max_combining_class = -1;
+        continue;
+      }
+      if (current_char >= 0x110000) {
+        /* out-of-range codepoint: drop it, as utf8proc_encode_char already
+           does (unsafe_get_property would OOB on utf8proc_stage1table) */
```
stevengj, line 676: Alternatively, should we just keep all invalid codepoints, merging with the `current_char < 0` branch?

apnadkarni: No expert in these matters, but invalid code points would result in UTF-8 representations longer than 4 bytes, something library clients may not expect. Dropping, afaik, is not recommended by the standard. Wouldn't replacement of invalid codepoints with U+FFFD be a preferable solution?

stevengj: Here we are talking about the `normalize_utf32` function, which uses UTF-32 for both input and output, and for UTF-32 there is no issue with overlong encodings.

It is a separate question of what to do when UTF-32 data is encoded into UTF-8.  Currently, it simply drops any invalid codepoints (< 0 or ≥ 0x110000), unless the `CHARBOUND` option is enabled (in which case `-1` codepoints are encoded as `0xFF` sentinel bytes).  That won't change (at least, not in this PR).

eschnett: I applied @stevengj's suggestion.

